PR #4, opened 2021-01-14
Got rid of v1 in module path to make it go-gettable

alex-ntousias:
# Title

## Type of change

## Description

Because we're still in version 1, we need to get rid of the version in the module path according to this: https://golang.org/ref/mod#major-version-suffixes

That was the reason I couldn't add the SDK as a dependency in the `clerk_go` project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,8 +9,7 @@ Go client library for accessing the [Clerk Server API v1](https://docs.clerk.dev
 First, add one of the following imports, depending on whether you use modules or not.
 
 ```go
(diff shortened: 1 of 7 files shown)


braden-clerk (approved):
well i finally learned why our go.mod  has a bunch of `+incompatible` at the end o version strings.  What a weird requirement

braden-clerk:
🙃  It’d be nice of our import statement read a bit cleaner, how hard would it be to make it
`import “github.com/clerkinc/clerk-sdk-go` ?

alex-ntousias:
@braden-clerk I'll see what I can do. We might need to rename the repository for that.

PR #5, opened 2021-01-18
Added remaining operations

alex-ntousias:
# Title

## Type of change

## Description

Quite a few changes in this PR:
* Added client operations
* Updated email, sms and user endpoints according to the latest server api changes
* Added verification operation
* Added Clerk SDK in `pkg.go.dev`

### Related Issue (optional)

https://www.notion.so/clerkdev/Server-API-Go-SDK-7b90fb68036e41e8afd60366b9a774e2


clerk/clients.go
```diff
@@ -0,0 +1,49 @@
...
+	var clients []ClientResponse
+	_, err := s.client.Do(req, &clients)
+	if err != nil {
+		return nil, err
+	}
+	return clients, nil
+}
+
+func (s *ClientsService) Read(clientId string) (*ClientResponse, error) {
+	clientUrl := fmt.Sprintf("clients/%v", clientId)
```
SokratisVidros, line 29: 🙃 Do we still need this? If no please remove.

clerk/verification.go
```diff
@@ -0,0 +1,72 @@
...
+	OriginHeader        = "Origin"
+)
+
+type VerificationService service
+
+type verifyRequest struct {
+	Token string `json:"token"`
+}
+
+func (s *VerificationService) Verify(req *http.Request) (*Session, error) {
```
SokratisVidros, line 21: ⛔ I'd suggest that we also export a pure plug n play, HTTP Go middleware too. For example

```go
func clerkMiddleware(next http.Handler) http.Handler {
  return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
    doClerkStuff()...
    next.ServeHTTP(w, r)  })
}
```

WDYT?

SokratisVidros (changes requested):

clerk/verification.go
alex-ntousias: Yes, I was planning to add something like that. 
I just wanted to first start using it, to make sure I implement it according to an actual use case. But I can also add it now and modify it accordingly once we include it in the DApi.

SokratisVidros: 👍

SokratisVidros (approved):
🎖️

clerk/clients.go
alex-ntousias: Not needed for the verification, but I can still see this operation in the server API. 
Are we planning to remove it?

SokratisVidros: I was referring to line 29 that prints the ClientURL

alex-ntousias: Ah, that's not a print. 
It's for formatting the string, basically similar to `"clients/" + clientId`.

SokratisVidros: Ouch, that's right. My bad. Apologies.

clerk/verification.go
```diff
@@ -0,0 +1,72 @@
...
+
+	if sessionId == "" {
+		return nil, errors.New(fmt.Sprintf("no session id is specified via the %s query parameter", QueryParamSessionId))
+	}
+
+	return s.client.Sessions().Verify(sessionId, sessionToken)
+}
+
+func isXHR(req *http.Request) bool {
+	headers := req.Header
```
agis, line 45: 💡 For being more future-proof, we could also check here that `req` is not nil. I know that currently only `Verify` calls into this and this is redundant, but that could change in the future.

SokratisVidros (approved):

clerk/verification.go
```diff
@@ -0,0 +1,72 @@
...
+	origin := headers.Get(OriginHeader)
+	return len(origin) > 0
```
agis, line 47: ❓ Is this better than checking for the `X-Requested-With` header? Or maybe we could use both?

clerk/verification.go
```diff
@@ -0,0 +1,72 @@
...
+
+func doVerify(client Client, url string, token string, response interface{}) error {
+	tokenPayload := verifyRequest{Token: token}
+	req, _ := client.NewRequest("POST", url, &tokenPayload)
+
+	_, err := client.Do(req, response)
+	if err != nil {
+		return err
+	}
+	return nil
```
agis, line 71: 🔧 You can simply do:

```golang
return client.Do(req, response)
```

clerk/clients.go
alex-ntousias: @agis 👍 Sounds good.

clerk/clerk.go
```diff
@@ -21,8 +21,10 @@ type Client interface {
 
 	Users() *UsersService
 	Sessions() *SessionsService
+	Clients() *ClientsService
 	Emails() *EmailService
 	SMS() *SMSService
+	Verification() *VerificationService
```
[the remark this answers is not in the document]
alex-ntousias: 👍

clerk/verification.go
agis: Nevermind my previous comment, I meant to propose this:

```golang
_, err := client.Do(req, response)
return err
```

alex-ntousias: Ah, good catch!

README.md
```diff
@@ -1,12 +1,19 @@
...
 [![Test Status](https://github.com/clerkinc/clerk-sdk-go/workflows/tests/badge.svg)](https://github.com/clerkinc/clerk-sdk-go/actions?query=workflow%3Atests)
 
 Go client library for accessing the [Clerk Server API v1](https://docs.clerk.dev/server-api/).
 
 ## Usage ##
 
-First, add one of the following imports, depending on whether you use modules or not.
+First, add the Clerk SDK as a dependency to your project.
+
+```go
```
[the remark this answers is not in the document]
alex-ntousias: 👍

clerk/clients.go
```diff
@@ -0,0 +1,49 @@
...
+	var clientResponse ClientResponse
+	_, err := s.client.Do(req, &clientResponse)
+	if err != nil {
+		return nil, err
+	}
+	return &clientResponse, nil
+}
+
+func (s *ClientsService) Verify(token string) (*ClientResponse, error) {
+	verifyUrl := "clients/verify"
```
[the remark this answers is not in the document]
alex-ntousias: Nice!

clerk/emails.go
```diff
@@ -1,31 +1,25 @@
...
+	ToEmailAddress *string `json:"to_email_address,omitempty"`
 	Email
 }
 
 // Sends an email.
 // Please note that one of ToEmailAddress or EmailAddressID must be supplied.
 func (s *EmailService) Create(email Email) (*EmailResponse, error) {
-	if email.ToEmailAddress == nil && email.EmailAddressID == nil {
-		return nil, errors.New("one of ToEmailAddress or EmailAddressID must be supplied")
-	}
```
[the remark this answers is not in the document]
alex-ntousias: That's why I don't like comments as documentation. They become out-of-sync soooo fast! Very good catch!

clerk/verification.go
alex-ntousias: No idea to be honest! Let's consult with Colin (who came up with this algorithm) later today and modify it accordingly.

clerk/verification.go
alex-ntousias: I'm not a huge fan of null-checks and error checks everywhere. 

I believe that we should be defensive and make the necessary checks on the boundaries of our application (where we're dealing with external inputs) but other than that, trust everything else that is in our control.

clerk/emails.go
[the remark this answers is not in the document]
agis: Yeah that particular one should be removed completely imho

clerk/clients.go
```diff
@@ -0,0 +1,49 @@
...
+	_, err := s.client.Do(req, &clients)
+	if err != nil {
+		return nil, err
+	}
+	return clients, nil
+}
+
+func (s *ClientsService) Read(clientId string) (*ClientResponse, error) {
+	clientUrl := fmt.Sprintf("clients/%v", clientId)
+	req, _ := s.client.NewRequest("GET", clientUrl)
```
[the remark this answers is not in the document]
alex-ntousias: The `Client.NewRequest` can return errors if we're trying to create an invalid request, e.g. invalid request method.

Since, at this point, we are in control of what we are creating, we cannot have an error.

We could, of course, have an error check here. But as I mentioned in the [other comment](https://github.com/clerkinc/clerk-sdk-go/pull/5#discussion_r560047293), I'm not a huge fan of defensive programming inside the boundaries of our application. 

I'm actually an advocate of the (arguably) radical philosophy which says that anything that can't be tested should be removed. That's one of the reasons that I wanted to have 100% coverage on this...it doesn't help you eliminate the bugs but it helps identifying cases that can never occur.

README.md
```diff
@@ -1,12 +1,19 @@
...
 
+[![GoDoc](https://img.shields.io/static/v1?label=godoc&message=reference&color=blue)](https://pkg.go.dev/github.com/clerkinc/clerk-sdk-go/clerk)
 [![Test Status](https://github.com/clerkinc/clerk-sdk-go/workflows/tests/badge.svg)](https://github.com/clerkinc/clerk-sdk-go/actions?query=workflow%3Atests)
 
 Go client library for accessing the [Clerk Server API v1](https://docs.clerk.dev/server-api/).
 
 ## Usage ##
 
-First, add one of the following imports, depending on whether you use modules or not.
+First, add the Clerk SDK as a dependency to your project.
```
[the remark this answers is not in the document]
alex-ntousias: Oh, I actually thought that the `go get` command is how you add dependencies in your project when using modules.
Isn't that the case?

Sorry in advance if these questions are a bit stupid, I'm quite new in Go (actually this project is the first one I've used the language).

clerk/clerk.go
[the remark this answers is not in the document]
SokratisVidros: Yes! OCD FTW 🤣 🤣 🤣

README.md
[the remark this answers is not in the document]
agis: Not stupid at all! In fact, you're right, `go mod` does in fact automatically adds the module to `go.mod` - I was under the impression that it didn't.

FWIW [here](https://github.com/golang/go/issues/27643) is a relevant issue where this is discussed.

clerk/verification.go
agis: cc @colinclerk 

Truth is, you can't say with absolute certainty whether a request is an XHR so this is anyway going to be a best-effort approach. I'm just wondering if `X-Requested-With` is the better way of doing this.

colinclerk: @agis is right, relying on Origin isn't safe.  Apparently, the Origin is not sent with same-origin requests, and since the SDK handles requests from the developers frontend to the developers backend, it's completely possible the request will be same-origin.

It doesn't look like `X-Requested-With` is safe either, though.  It seems some frameworks may add `X-Requested-With`, but raw fetch and XHR do not.

I put up a proposal for resolution here: https://docs.google.com/document/d/19aUhuOR_zIDti7yWafkGw3wMG026guAObvM_yHe80UM/edit?usp=sharing

PR #12, opened 2021-04-22
If no session id is given, get the latest active from client's sessions

alex-ntousias:
## Type of change

## Description

Given that most applications will be single-session, we had to simplify the verification process for cases where we were not given a session id.

We modified the response of `/clients/verify` endpoint to return the list of sessions along with the client.
So, we don't need to have an additional call to the backend API anymore to get active session for the client, we can just get it from the returned sessions.

### Related Issue (optional)

https://www.notion.so/clerkdev/In-single-session-mode-without-_clerk_session_id-come-up-with-strategy-to-use-1-api-request-or-0--ab0fac16668c4dd487f668bd30077e4d

### DIFF
--- a/clerk/verification.go
+++ b/clerk/verification.go
@@ -45,7 +45,13 @@ func (s *VerificationService) useClientActiveSession(token string) (*Session, er
 		return nil, errors.New("no active sessions for given client")
 	}
 
-	return s.client.Sessions().Read(*clientResponse.LastActiveSessionID)
+	for _, session := range clientResponse.Sessions {
+		if session.ID == *clientResponse.LastActiveSessionID {
+			return session, nil
+		}
+	}
+
+	return nil, errors.New("active session not included in client's sessions")
 }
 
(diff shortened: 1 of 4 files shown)


clerk/verification.go
```diff
@@ -45,7 +45,19 @@ func (s *VerificationService) useClientActiveSession(token string) (*Session, er
...
+			lastActiveSession = session
+			break
+		}
+	}
+
+	if lastActiveSession == nil {
+		return nil, errors.New("active session not included in client's sessions")
+	}
+
+	return lastActiveSession, nil
```
chanioxaris, line 60: This can be simplified as below without the need of new variable initialization.

```
for _, session := range clientResponse.Sessions {
		if session.ID == *clientResponse.LastActiveSessionID {
                         return session, nil
		}
}

return nil, errors.New("active session not included in client's sessions")
```

chanioxaris (approved):
Nice work 💯

PR #20, opened 2021-07-30
feat: Introduce DecodeToken and VerifyToken utilities

chanioxaris:
## Type of change

## Description

This PR adds two function utilities in the Go SDK to be used with the upcoming networkless authentication. 
In particular, it adds the following:

- DecodeToken: Return the token claims included in the JWT token **without** verifying it
- VerifyToken: Verify the JWT token and return the token claims

Must be merged after #19

### DIFF
--- a/clerk/clerk.go
+++ b/clerk/clerk.go
@@ -26,6 +26,9 @@ type Client interface {
 	NewRequest(method string, url string, body ...interface{}) (*http.Request, error)
 	Do(req *http.Request, v interface{}) (*http.Response, error)
 
+	DecodeToken(token string) (*Claims, error)
+	VerifyToken(token string) (*Claims, error)
+
 	Clients() *ClientsService
 	Emails() *EmailService
 	JWKS() *JWKSService
@@ -41,9 +44,10 @@ type service struct {
 }
 
 type client struct {
-	client  *http.Client
-	baseURL *url.URL
-	token   string
+	client    *http.Client
+	baseURL   *url.URL
+	jwksCache *jwksCache
+	token     string
 
 	clients      *ClientsService
 	emails       *EmailService
@@ -81,6 +85,8 @@ func NewClientWithCustomHTTP(token string, urlStr string, httpClient *http.Clien
 	client.users = (*UsersService)(commonService)
 	client.webhooks = (*WebhooksService)(commonService)
 	client.verification = (*VerificationService)(commonService)
+
+	client.jwksCache = &jwksCache{}
 
 	return client, nil
 }
--- a/clerk/jwks_cache.go
+++ b/clerk/jwks_cache.go
@@ -1,0 +1,43 @@
+package clerk
+
+import (
+	"fmt"
+	"sync"
+	"time"
+
+	"gopkg.in/square/go-jose.v2"
+)
+
+type jwksCache struct {
+	sync.RWMutex
+	jwks      *JWKS
+	expiresAt time.Time
+}
+
+func (c *jwksCache) isInvalid() bool {
+	c.RLock()
+	defer c.RUnlock()
+
+	return c.jwks == nil || len(c.jwks.Keys) == 0 || time.Now().After(c.expiresAt)
+}
+
+func (c *jwksCache) set(jwks *JWKS) {
+	c.Lock()
+	defer c.Unlock()
+
+	c.jwks = jwks
+	c.expiresAt = time.Now().Add(time.Hour)
+}
+
+func (c *jwksCache) get(kid string) (*jose.JSONWebKey, error) {
+	c.RLock()
+	defer c.RUnlock()
+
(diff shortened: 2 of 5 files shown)


clerk/tokens.go
```diff
@@ -0,0 +1,100 @@
...
+	claims := Claims{}
+	if err = parsedToken.UnsafeClaimsWithoutVerification(&claims); err != nil {
+		return nil, err
+	}
+
+	return &claims, nil
+}
+
+// VerifyToken verifies the session jwt token.
+func (c *client) VerifyToken(token string) (*Claims, error) {
```
gkats, line 48: I see the Ruby SDK accepts some additional parameters. https://github.com/clerkinc/clerk-sdk-ruby/blob/main/lib/clerk/sdk.rb#L151

Is it worth it to try something similar?

cc @agis @SokratisVidros

clerk/tokens.go
```diff
@@ -0,0 +1,100 @@
...
+	PreferredUsername   string `json:"preferred_username,omitempty"`
+	Gender              string `json:"gender,omitempty"`
+	Birthdate           string `json:"birthdate,omitempty"`
+	Email               string `json:"email,omitempty"`
+	EmailVerified       bool   `json:"email_verified,omitempty"`
+	PhoneNumber         string `json:"phone_number,omitempty"`
+	PhoneNumberVerified bool   `json:"phone_number_verified,omitempty"`
+}
+
+// DecodeToken decodes the session jwt token.
```
gkats, line 32: :upside_down_face: Perhaps we can add something in the godoc about the token not being verified, just decoded?

clerk/jwks_cache.go
```diff
@@ -0,0 +1,32 @@
...
+}
+
+func (c *jwksCache) set(jwks *JWKS) {
+	c.jwks = jwks
+	c.expiresAt = time.Now().Add(time.Hour)
+}
+
+func (c jwksCache) get(kid string) (*jose.JSONWebKey, error) {
+	for _, key := range c.jwks.Keys {
+		if key.KeyID == kid {
```
SokratisVidros, line 26: 🔧 We should raise an error if there is no match to fail fast.

SokratisVidros (changes requested):

clerk/tokens.go
chanioxaris: Regarding the timeout, someone can set it by passing a custom http client in the sdk constructor `NewClientWithCustomHTTP()`. Although i think we should consider passing a context as the first argument to every function that has to deal with external http request. In this way we will give the developer the option to handle timeout/cancel independently for every operation. 
For the rest options and taking into account Go's way to deal with optional parameters, i don't think its worth at the current state

clerk/jwks_cache.go
chanioxaris: We have to iterate first over the keys to check if there is a match with the provided kid

clerk/jwks_cache.go
```diff
@@ -0,0 +1,32 @@
+package clerk
+
+import (
+	"fmt"
+	"time"
+
+	"gopkg.in/square/go-jose.v2"
+)
+
+type jwksCache struct {
```
agis, line 10: 🔧 Since this cache might be called from different goroutines concurrently, it should be made goroutine-safe to prevent data races.

SokratisVidros (approved):

clerk/clerk.go
```diff
@@ -26,6 +26,9 @@ type Client interface {
 	NewRequest(method string, url string, body ...interface{}) (*http.Request, error)
 	Do(req *http.Request, v interface{}) (*http.Response, error)
 
+	DecodeToken(token string) (*Claims, error)
+	VerifyToken(token string) (*Claims, error)
```
agis, line 30: Should we perhaps document those? We should explicitly state that decode does not verify the token.

clerk/jwks_cache.go
```diff
@@ -0,0 +1,40 @@
...
+	sync.Mutex
+	jwks      *JWKS
+	expiresAt time.Time
+}
+
+func (c *jwksCache) isInvalid() bool {
+	return c.jwks == nil || len(c.jwks.Keys) == 0 || time.Now().After(c.expiresAt)
```
agis, line 18: This call has to be guarded as well.

clerk/jwks_cache.go
```diff
@@ -0,0 +1,40 @@
...
+import (
+	"fmt"
+	"sync"
+	"time"
+
+	"gopkg.in/square/go-jose.v2"
+)
+
+type jwksCache struct {
+	sync.Mutex
```
agis, line 12: I'd recommend to use a `sync.RWMutex` instead, since we want to be able to have multiple goroutines reading this value simultaneously.

chanioxaris: Nice catch

PR #21, opened 2021-08-19
feat: Introduce middleware for networkless authentication

chanioxaris:
## Type of change

## Description

This PR introduces a new middleware `WithNetworkless()` for the upcoming networkless authentication that retrieves a JWT token from the `Authorization` header and tries to verify it. If successful adds the JWT claims in the request context so you can easily access them inside a handler

### DIFF
--- a/clerk/middleware.go
+++ b/clerk/middleware.go
@@ -2,11 +2,13 @@ package clerk
 
 import (
 	"context"
+	"errors"
 	"net/http"
 )
 
 const (
 	ActiveSession = iota
+	ActiveClaims
 )
 
 func WithSession(client Client) func(handler http.Handler) http.Handler {
@@ -29,4 +31,42 @@ func addSessionToContext(request *http.Request, session *Session) *http.Request 
 	ctx := request.Context()
 	updated := context.WithValue(ctx, ActiveSession, session)
 	return request.WithContext(updated)
+}
+
+func WithNetworkless(client Client) func(handler http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			token, err := retrieveTokenFromRequest(r)
+			if err != nil {
+				w.WriteHeader(http.StatusUnauthorized)
+				_, _ = w.Write([]byte(err.Error()))
+				return
+			}
+
+			claims, err := client.VerifyToken(token)
+			if err != nil {
+				w.WriteHeader(http.StatusUnauthorized)
+				_, _ = w.Write([]byte(err.Error()))
+				return
+			}
+
+			ctx := context.WithValue(r.Context(), ActiveClaims, claims)
+			next.ServeHTTP(w, r.WithContext(ctx))
+		})
+	}
+}
+
+func retrieveTokenFromRequest(r *http.Request) (string, error) {
+	token := r.Header.Get("Authorization")
+	if token != "" {
+		return token, nil
+	}
+
+	for _, cookie := range r.Cookies() {
+		if cookie.Name == "__session" {
+			return cookie.Value, nil
+		}
+	}
+
+	return "", errors.New("no token found in request header or cookie")
 }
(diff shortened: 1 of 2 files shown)


clerk/middleware.go
```diff
@@ -30,3 +31,21 @@ func addSessionToContext(request *http.Request, session *Session) *http.Request
 	updated := context.WithValue(ctx, ActiveSession, session)
 	return request.WithContext(updated)
 }
+
+func WithNetworkless(client Client) func(handler http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			token := r.Header.Get("Authorization")
+
```
agis, line 39: ❓ I think it'll be better DX to check here if the header is empty and if it is, return a respective error indicating the issue.

clerk/middleware.go
```diff
@@ -30,3 +31,21 @@ func addSessionToContext(request *http.Request, session *Session) *http.Request
 	updated := context.WithValue(ctx, ActiveSession, session)
 	return request.WithContext(updated)
 }
+
+func WithNetworkless(client Client) func(handler http.Handler) http.Handler {
```
agis, line 35: 🔧 Perhaps, for consistency, `WithSessionNetworkless` is a better name?

clerk/middleware.go
```diff
@@ -30,3 +31,21 @@ func addSessionToContext(request *http.Request, session *Session) *http.Request
...
+			if err != nil {
+				w.WriteHeader(http.StatusUnauthorized)
+				_, _ = w.Write([]byte(err.Error()))
+				return
+			}
+
+			ctx := context.WithValue(r.Context(), ActiveClaims, claims)
+			next.ServeHTTP(w, r.WithContext(ctx))
+		})
+	}
```
agis, line 50: Apart from the `Authorization` header, we additionally have to try and fetch the token from the `__session` cookie, based on architecture no. 1, right?

clerk/middleware.go
alex-ntousias: I agree with @agis about the extra-check about the token being empty. But I would choose to place this check into the `VerifyToken` function since we already have a number of checks there regarding the token and the `VerifyToken` can also be used directly (without having to go via the middleware).

clerk/middleware.go
```diff
@@ -30,3 +31,21 @@ func addSessionToContext(request *http.Request, session *Session) *http.Request
...
 }
+
+func WithNetworkless(client Client) func(handler http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			token := r.Header.Get("Authorization")
+
+			claims, err := client.VerifyToken(token)
+			if err != nil {
+				w.WriteHeader(http.StatusUnauthorized)
```
alex-ntousias, line 42: Is the status always going to be `401 Unauthorized`  for every error returned by the `VerifyToken`?

alex-ntousias, line 42: Is the status always going to be `401 Unauthorized` for every error returned by the `VerifyToken`?

alex-ntousias (commented):
LGTM @chanioxaris 👍

clerk/middleware.go
chanioxaris: I believe so, if something went wrong during token verification i think the proper response status code is 401

clerk/middleware.go
chanioxaris: Good catch, it's possible someone to serve a UI via a Go binary

clerk/middleware.go
chanioxaris: Thank you guys for pointing this out. I will also handle the case the token is empty inside the middleware to early return

clerk/middleware.go
chanioxaris: Not sure about the name yet, although this will be an easy change

SokratisVidros (commented):
I agree with all the above comments. Naming is still a pending issue that we will discuss in about 20 minutes.

PR #22, opened 2021-09-02
refactor: DecodeToken to be generic and VerifyToken to use new claims

chanioxaris:
## Type of change

## Description

This PR changes the `DecodeToken` utility to decode any jwt token and not only a specific. Also updated the claims that the session token will include accordingly within the `VerifyToken` utility

### DIFF
--- a/clerk/tokens.go
+++ b/clerk/tokens.go
@@ -2,6 +2,7 @@ package clerk
 
 import (
 	"fmt"
+	"strings"
 	"time"
 
 	"gopkg.in/square/go-jose.v2"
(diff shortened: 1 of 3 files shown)


clerk/tokens.go
```diff
@@ -69,18 +71,17 @@ func (c *client) VerifyToken(token string) (*Claims, error) {
...
-	expectedClaims := jwt.Expected{
-		Audience: jwt.Audience{"clerk"},
-		Time:     time.Now(),
+	if err = claims.Claims.ValidateWithLeeway(jwt.Expected{Time: time.Now()}, 0); err != nil {
+		return nil, err
 	}
 
-	if err = claims.Claims.ValidateWithLeeway(expectedClaims, 0); err != nil {
-		return nil, err
+	if !strings.HasPrefix(claims.Issuer, "clerk.") {
```
SokratisVidros, line 83: 🔧 Shouldn't this be

```suggestion
	if !strings.HasPrefix(claims.Issuer, "https://clerk.") {
```

SokratisVidros (approved):

clerk/tokens.go
chanioxaris: Yes thats correct, i changed it in the middleware PR #23, although i will fix it here as it's the proper place

PR #23, opened 2021-09-02
feat: Change middleware to also handle AuthV2

chanioxaris:
## Type of change

## Description

This PR changes the current middleware `WithSession` to also handle session tokens that will be used in the upcoming networkless authentication. If successful adds the JWT claims in the request context so you can easily access them inside a handler. This change should not affect or change the middleware behavior for the current use cases

### DIFF
--- a/clerk/clerk.go
+++ b/clerk/clerk.go
@@ -26,8 +26,8 @@ type Client interface {
 	NewRequest(method string, url string, body ...interface{}) (*http.Request, error)
 	Do(req *http.Request, v interface{}) (*http.Response, error)
 
-	DecodeToken(token string) (*Claims, error)
-	VerifyToken(token string) (*Claims, error)
+	DecodeToken(token string) (*TokenClaims, error)
+	VerifyToken(token string) (*SessionClaims, error)
 
 	Clients() *ClientsService
 	Emails() *EmailService
--- a/clerk/middleware.go
+++ b/clerk/middleware.go
@@ -3,30 +3,63 @@ package clerk
 import (
 	"context"
 	"net/http"
+	"strings"
 )
 
 const (
 	ActiveSession = iota
+	ActiveSessionClaims
 )
 
 func WithSession(client Client) func(handler http.Handler) http.Handler {
 	return func(next http.Handler) http.Handler {
-		return http.HandlerFunc(func(writer http.ResponseWriter, request *http.Request) {
-			session, err := client.Verification().Verify(request)
-			if err != nil {
-				writer.WriteHeader(400)
-				writer.Write([]byte(err.Error()))
-				return
-			}
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			if token, isAuthV2 := isAuthV2Request(r, client); isAuthV2 {
+				// Validate using session token
+				claims, err := client.VerifyToken(token)
+				if err != nil {
+					w.WriteHeader(http.StatusUnauthorized)
+					_, _ = w.Write([]byte(err.Error()))
+					return
+				}
 
-			updatedRequest := addSessionToContext(request, session)
-			next.ServeHTTP(writer, updatedRequest)
+				ctx := context.WithValue(r.Context(), ActiveSessionClaims, claims)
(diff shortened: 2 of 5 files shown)


clerk/middleware.go
```diff
@@ -3,30 +3,63 @@ package clerk
...
+					return
+				}
+
+				ctx := context.WithValue(r.Context(), ActiveSessionClaims, claims)
+				next.ServeHTTP(w, r.WithContext(ctx))
+			} else {
+				// Validate using session verify request
+				session, err := client.Verification().Verify(r)
+				if err != nil {
+					w.WriteHeader(http.StatusUnauthorized)
```
chanioxaris, line 32: Before we returned a 400 Bad Request if the verification failed. Do you think we should change it to a 401 Unauthorized as more appropriate status code?

clerk/clerk.go
```diff
@@ -26,8 +26,8 @@ type Client interface {
 	NewRequest(method string, url string, body ...interface{}) (*http.Request, error)
 	Do(req *http.Request, v interface{}) (*http.Response, error)
 
-	DecodeToken(token string) (*Claims, error)
-	VerifyToken(token string) (*Claims, error)
+	DecodeToken(token string) (*TokenClaims, error)
+	VerifyToken(token string) (*SessionClaims, error)
```
agis, line 30: ❓ Isn't this a breaking change? Or do these new types have the same API as the previous one? (Even so, it's still breaking for users that checked the actual type).

chanioxaris: After we add these function utilities, we haven't released a new sdk version so i doubt anyone are using them at all

agis: Yeah forgot that 😊

clerk/tokens.go
```diff
@@ -2,50 +2,52 @@ package clerk
...
-	FamilyName          string `json:"family_name,omitempty"`
-	PreferredUsername   string `json:"preferred_username,omitempty"`
-	Gender              string `json:"gender,omitempty"`
-	Birthdate           string `json:"birthdate,omitempty"`
-	Email               string `json:"email,omitempty"`
-	EmailVerified       bool   `json:"email_verified,omitempty"`
-	PhoneNumber         string `json:"phone_number,omitempty"`
-	PhoneNumberVerified bool   `json:"phone_number_verified,omitempty"`
+type SessionClaims struct {
+	jwt.Claims
```
agis, line 23: 🔧 Why don't we specify explicitly only the ones that the session tokens contain?

clerk/tokens.go
```diff
@@ -2,50 +2,52 @@ package clerk
...
 
-type Claims struct {
+var (
+	standardClaimsKeys = []string{"iss", "sub", "aud", "exp", "nbf", "iat", "jti"}
+)
+
+type TokenClaims struct {
 	jwt.Claims
-	UserClaims
+	Extra map[string]interface{}
```
agis, line 19: 🔧 Instead of using an extra field, can't we just have this be a `map`? Since this is for decoding any JWT really..

chanioxaris: I decided to embed the standard claims, just to make users life easier so they don't need to check if exists in the map

clerk/tokens.go
chanioxaris: Will change it accordingly to only contain the session token claims

chanioxaris: Actually by not using the `jwt.Claims` we will have to implement the validation on our own and we won't leverage the package's helper method `ValidateWithLeeway`

agis: Cool, let's leave it as is then.

clerk/tokens.go
agis: Fair point!

clerk/tokens_test.go
```diff
@@ -16,19 +16,38 @@ import (
 )
 
 var (
-	dummyClaims = Claims{
+	dummyTokenClaims = map[string]interface{}{
+		"iss":     "issuer",
+		"sub":     "subject",
+		"aud":     []string{"clerk"},
```
agis, line 22: Remove this

agis (approved):

PR #36, opened 2021-12-07
fix: Templates integration tests to include the required variables per slug

chanioxaris:
## Type of change

## Description

After the recent changes in the templates endpoint, we need to include the required variables for each template slug, otherwise the integration tests will fail, blocking any new PR


yourtallness (approved):
:shipit:

Issue #35, opened 2021-12-07
V2 middleware does not permit requests without authentication


notproperlycut:
Hi, I've spoken with @SokratisVidros about this, who requested I add this issue.

We're trying to use your V2 auth middleware. One reason is so that we can use `WithSessionV2` to populate session claims into request context if we have a signed-in user, but permit the request even if we do not. The [README](https://github.com/clerkinc/clerk-sdk-go/blob/main/README.md) suggests that `RequireSessionV2` is for cases where one wants to be more strict about unauthorized requests.

In our instance, we are serving a graphQL endpoint behind `WithSessionV2`. For some requests, we do not want or need a signed-in user. For instance: we use [this npm package](https://github.com/prisma-labs/get-graphql-schema) to pull the schema from our endpoint, as part of our dev/build process.

Right now, completely header/cookie-free requests do not make it through the `WithSessionV2` handler. Instead we return some HTML designed to work with clerk-js and cause a retry. I might be wrong, but I believe it's [this line](https://github.com/clerkinc/clerk-sdk-go/blob/bf85e83891b0f02b673c7b619c20d62ef8bc5935/clerk/middleware_v2.go#L90) we are triggering.

This workflow worked for us previously, when we were using `client.Verification().Verify(r)` as described under "Auth v1 (deprecated)" [here](https://docs.clerk.dev/reference/backend-api-reference/sdks/golang/verifying-a-session). But since we moved to use the middleware, I don't see a way to let all requests through so that we can decide downstream, whether to 401/403

chanioxaris:
Hey there @notproperlycut, thanks for reporting this. 

From what you described i believe the SDK wrongfully renders an interstitial page (to load clerk.js as you stated) which is irrelevant to your case. 

Can you please share how you are testing this, where the backend and frontend runs (e.g. http://localhost:4000), so i can reproduce it on my own and push a fix?

notproperlycut:
Thanks @chanioxaris. The easiest way I found to trigger this condition, was to simply curl our endpoint running locally. So, with no frontend at all.

In my case...
```
$ curl http://localhost:8080/query              
<!DOCTYPE html>                                                                                                                                                                   
<html>                                                                                   
    <head>                                                                                                                                                                        
        <meta charset="UTF-8" />                                                         
    </head>                                                                              
    <body>                                                                               
        <script>                                                                                                                                                                  
            window.startClerk = async () => {                                            
                const Clerk = window.Clerk;
...
```

[...]

PR #37, opened 2021-12-08
feat: Support non-browser requests for development/staging environments

chanioxaris:
## Type of change

## Description

This PR contains the below changes:
1. For development/staging environments, try to detect if that's a non-browser request (e.g. script) based on the request `User Agent` and prevent interstitial rendering
2. Take into account the `Authorization` header every time it exists in the request

### Related Issue (optional)

#35

### DIFF
--- a/clerk/middleware_v2.go
+++ b/clerk/middleware_v2.go
@@ -43,17 +43,17 @@ func SessionFromContext(ctx context.Context) (*SessionClaims, bool) {
 func WithSessionV2(client Client) func(handler http.Handler) http.Handler {
 	return func(next http.Handler) http.Handler {
 		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-			headerToken := strings.TrimSpace(r.Header.Get("authorization"))
-			headerToken = strings.TrimPrefix(headerToken, "Bearer ")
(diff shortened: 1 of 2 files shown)


clerk/middleware_v2.go
```diff
@@ -73,6 +73,13 @@ func WithSessionV2(client Client) func(handler http.Handler) http.Handler {
 				return
 			}
 
+			// In development or staging environment and probably not a web browser request based on User Agent
+			if isDevelopmentOrStaging(client) && !strings.HasPrefix(r.UserAgent(), "Mozilla/") {
```
agis, line 77: I think we should explain *why* we do this instead of *how* here, considering it's a change in the core AuthV2 algorithm.

SokratisVidros (approved):

clerk/middleware_v2.go
chanioxaris: Added a brief explanation

agis:
@chanioxaris shall we include the fix for treating the empty auth header as logged out?

colinclerk (approved):
Thanks, lgtm!

SokratisVidros (approved):
🎖️ Looking good. Let's unit test this and transfer **the exact same logic to node and ruby SDK** on Monday.

PR #38, opened 2021-12-09
feat: Provide functional options for the client constructor

chanioxaris:
## Type of change

## Description

This PR introduces functional options for our client constructor, that allow to override some of the default settings. 
After this change we won't need to export multiple constructor functions (e.g. `clerk.NewClientWithBaseUrl()`), but a single one `clerk.NewClient()` which will accept some optional parameters.

For example if someone wants to provide their own http client

**Before:**
```golang
customHTTPClient := &http.Client{Timeout: time.Second * 30}

c, err := clerk.NewClientWithCustomHTTP("apiKey", clerk.ProdUrl, customHTTPClient)
```

**After:**
```golang
customHTTPClient := &http.Client{Timeout: time.Second * 30}

c, err := clerk.NewClient("apiKey", clerk.WithHTTPClient(customHTTPClient))
```

The old constructor functions are still present, to ensure backwards compatibility, but marked as deprecated and they will be removed in an upcoming major release

### DIFF
--- a/clerk/clerk_options.go
+++ b/clerk/clerk_options.go
@@ -1,0 +1,38 @@
+package clerk
+
+import (
+	"errors"
+	"net/http"
+)
+
+// ClerkOption describes a functional parameter for the clerk client constructor
+type ClerkOption func(*client) error
+
+// WithHTTPClient allows the overriding of the http client
+func WithHTTPClient(httpClient *http.Client) ClerkOption {
+	return func(c *client) error {
+		if httpClient == nil {
+			return errors.New("http client can't be nil")
+		}
+
+		c.client = httpClient
+		return nil
+	}
+}
+
+// WithBaseURL allows the overriding of the base URL
+func WithBaseURL(rawURL string) ClerkOption {
+	return func(c *client) error {
+		if rawURL == "" {
+			return errors.New("base url can't be empty")
+		}
+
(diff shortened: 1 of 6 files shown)


clerk/clerk_options.go
```diff
@@ -0,0 +1,37 @@
...
+		if httpClient == nil {
+			return errors.New("http client can't be nil")
+		}
+
+		c.client = httpClient
+		return nil
+	}
+}
+
+func WithBaseURL(rawURL string) ClientOption {
```
gkats, line 23: :upside_down_face: Let's add a godoc comment here as well.

gkats (approved):

clerk/clerk_options.go
```diff
@@ -0,0 +1,37 @@
+package clerk
```
agis, line 1: Maybe rename the file name to `client_options.go`, to match the actual type.

agis (approved):

clerk/clerk_options.go
chanioxaris: Nice catch missed that, thanks

clerk/clerk_options.go
chanioxaris: Actually renamed this from `ClientOption` to `ClerkOption`, so there is no confusion if this options are for the clerk client or the client resource

PR #43, opened 2022-01-27
feat: Support to verify authorized party (azp) session token claim

chanioxaris:
## Type of change

## Description

Added support to verify the authorized party (azp) claim, if it's included in the session token claims within the VerifyToken method and v2 middlewares

### DIFF
--- a/clerk/tokens.go
+++ b/clerk/tokens.go
@@ -21,7 +21,8 @@ type TokenClaims struct {
 
 type SessionClaims struct {
 	jwt.Claims
-	SessionID string `json:"sid"`
+	SessionID       string `json:"sid"`
+	AuthorizedParty string `json:"azp"`
 }
 
 // DecodeToken decodes a jwt token without verifying it.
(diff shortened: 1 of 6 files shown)


gkats (approved):

SokratisVidros:
@chanioxaris Is this added to node SDK as well? The only one left should be Ruby, right?

chanioxaris:
> @chanioxaris Is this added to node SDK as well? The only one left should be Ruby, right?

Not yet, it's under implementation

clerk/tokens.go
```diff
@@ -46,8 +47,18 @@ func (c *client) DecodeToken(token string) (*TokenClaims, error) {
 	return &TokenClaims{Claims: standardClaims, Extra: extraClaims}, nil
 }
 
+type verifyTokenOptions struct {
+	authorizedParties []string
```
agis, line 51: Can we make this a map so that we avoid the linear search?

chanioxaris: Nice catch, changed in [10dc7fa](https://github.com/clerkinc/clerk-sdk-go/pull/43/commits/10dc7fa305c1b2fffa2993a37f93b1f92461ece1)

PR #49, opened 2022-03-22
Added endpoint to get total count of users

alex-ntousias:
## Type of change

## Description

This PR includes two changes:
* Adds some missing params on the list users which are supported by our BAPI endpoint. Namely, it allows to filter users based on web3 wallets, usernames, user ids or a query. It also allows to order users by particular properties.
* Adds the new users count endpoint which return the total count of users that match the filtering params given. This is particularly helpful for callers who need to know the total count in order to properly use the limit/offset capabilities of our list users method. _Please note_ that this is a temporary solution until we properly implement pagination responses.

### Related Issue (optional)

https://www.notion.so/clerkdev/Return-user-total-count-in-BAPI-771264a4fccc461984f98003b0204499

### DIFF
--- a/tests/integration/users_test.go
+++ b/tests/integration/users_test.go
@@ -33,6 +33,14 @@ func TestUsers(t *testing.T) {
 		t.Fatalf("Users.ListAll returned nil")
 	}
 
+	userCount, err := client.Users().Count(clerk.ListAllUsersParams{})
+	if err != nil {
+		t.Fatalf("Users.Count returned error: %v", err)
+	}
+	if userCount.TotalCount != len(users) {
+		t.Fatalf("Users.Count returned %d, expected %d", userCount.TotalCount, len(users))
+	}
+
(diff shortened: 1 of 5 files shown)


tests/integration/users_test.go
```diff
@@ -33,6 +33,14 @@ func TestUsers(t *testing.T) {
 		t.Fatalf("Users.ListAll returned nil")
 	}
 
+	userCount, err := client.Users().Count(clerk.ListAllUsersParams{})
+	if err != nil {
+		t.Fatalf("Users.Count returned error: %w", err)
```
chanioxaris, line 38:
```suggestion
		t.Fatalf("Users.Count returned error: %v", err)
```

chanioxaris (approved):

tests/integration/users_test.go
alex-ntousias: 👍

PR #59, opened 2022-07-28
fix: If the request contains a body, include content type header

alex-ntousias:
## Type of change

## Description

If the request we'll be sending contains a body, we should include the `Content-Type` header with value `application/json`.

### DIFF
--- a/clerk/clerk.go
+++ b/clerk/clerk.go
@@ -159,7 +159,9 @@ func (c *client) NewRequest(method, url string, body ...interface{}) (*http.Requ
 	}
 
 	var buf io.ReadWriter
+	var hasBody bool
 	if len(body) > 0 && body[0] != nil {
+		hasBody = true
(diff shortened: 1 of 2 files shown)


SokratisVidros (approved):
@alex-ntousias Is this the only change we need to do in order to support application/json requests?

Can you please make sure that endpoints who use nested JSON payloads work as expected. For example user and org metadata endpoints.

alex-ntousias:
@SokratisVidros Unfortunately, having a closer look at our BAPI request handling, this change doesn't look that safe any more! 
We have some special logic there to handle calls from our Go SDK, so that we convert the request to form content type. Unfortunately, this is necessary at the moment, because we need to support the NodeSDK which still sends form instead of JSON AFAIK.

So, I'm closing this PR and we'll revisit that when we work on BAPI v2.

PR #91, opened 2023-01-12
feat: Allow organizations update to receive metadata

AlexNti:
## Type of change

## Description

The UpdateMetadata function is only allow us to append or modify the params, the deletions can be done only by setting the value of a property to null. This adds extra complexity to the front end in order to perform deletion. We modify the Update organizations function in order to receive metadata so we will be able to delete easier.


alex-ntousias (approved):
🚀 🚀 🚀

PR #94, opened 2023-01-23
feat: Add get organization memeberships function

AlexNti:

## Description

Introduce a new function that will allow us to retrieve the memberships of an organization.

### DIFF
--- a/clerk/organizations.go
+++ b/clerk/organizations.go
@@ -161,3 +161,56 @@ func (s *OrganizationsService) CreateInvitation(params CreateOrganizationInvitat
 	_, err := s.client.Do(req, &organizationInvitation)
 	return &organizationInvitation, err
 }
+
+type ListOrganizationMembershipsParams struct {
+	OrganizationID string
+	Limit          *int
+	Offset         *int
+}
+
+type ΟrganizationMembershipPublicUserData struct {
+	FirstName       *string `json:"first_name"`
+	LastName        *string `json:"last_name"`
+	ProfileImageURL string  `json:"profile_image_url"`
+	Identifier      string  `json:"identifier"`
+	UserID          string  `json:"user_id"`
+}
+
+type OrganizationMembership struct {
+	Object          string          `json:"object"`
+	ID              string          `json:"id"`
+	PublicMetadata  json.RawMessage `json:"public_metadata"`
+	PrivateMetadata json.RawMessage `json:"private_metadata"`
+	Role            string          `json:"role"`
+	CreatedAt       int64           `json:"created_at"`
+	UpdatedAt       int64           `json:"updated_at"`
+
+	Organization   *Organization                         `json:"organization"`
+	PublicUserData *ΟrganizationMembershipPublicUserData `json:"public_user_data"`
+}
+
+type ListOrganizationMembershipsResponse struct {
+	Data       []OrganizationMembership `json:"data"`
+	TotalCount int64                    `json:"total_count"`
+}
+
+func (s *OrganizationsService) ListMemberships(params ListOrganizationMembershipsParams) (*ListOrganizationMembershipsResponse, error) {
+	endpoint := fmt.Sprintf("%s/%s/memberships", OrganizationsUrl, params.OrganizationID)
+	req, _ := s.client.NewRequest(http.MethodGet, endpoint)
+
+	query := req.URL.Query()
+	if params.Limit != nil {
+		query.Set("limit", strconv.Itoa(*params.Limit))
+	}
+	if params.Offset != nil {
+		query.Set("offset", strconv.Itoa(*params.Offset))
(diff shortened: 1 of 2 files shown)


clerk/organizations.go
```diff
@@ -161,3 +161,60 @@ func (s *OrganizationsService) CreateInvitation(params CreateOrganizationInvitat
...
+	FirstName       *string `json:"first_name"`
+	LastName        *string `json:"last_name"`
+	ProfileImageURL string  `json:"profile_image_url"`
+	Identifier      string  `json:"identifier"`
+}
+
+type organizationPublicUserData struct {
+	publicUserData
+	UserID string `json:"user_id"`
+}
```
gkats, line 181: :wrench: Why are there two structs instead of just one? What's the purpose of splitting this struct into `publicUserData` and the `UserID`?

clerk/organizations.go
```diff
@@ -161,3 +161,60 @@ func (s *OrganizationsService) CreateInvitation(params CreateOrganizationInvitat
...
+	Object          string          `json:"object"`
+	ID              string          `json:"id"`
+	PublicMetadata  json.RawMessage `json:"public_metadata"`
+	PrivateMetadata json.RawMessage `json:"private_metadata,omitempty"`
+	Role            string          `json:"role"`
+	CreatedAt       int64           `json:"created_at"`
+	UpdatedAt       int64           `json:"updated_at"`
+
+	Organization   *Organization               `json:"organization"`
+	PublicUserData *organizationPublicUserData `json:"public_user_data,omitempty"`
```
gkats, line 193: :wrench: If we're returning the `PublicUserData` then we should also export the type. Something like `clerk.OrganizationPublicUserData` or `clerk.OrganizationMembershipPublicUserData`. 

These will never be `null`, so the field should not be a pointer and there's no need to `omitempty` here.

clerk/organizations.go
```diff
@@ -161,3 +161,60 @@ func (s *OrganizationsService) CreateInvitation(params CreateOrganizationInvitat
...
+type organizationPublicUserData struct {
+	publicUserData
+	UserID string `json:"user_id"`
+}
+
+type OrganizationMembership struct {
+	Object          string          `json:"object"`
+	ID              string          `json:"id"`
+	PublicMetadata  json.RawMessage `json:"public_metadata"`
+	PrivateMetadata json.RawMessage `json:"private_metadata,omitempty"`
```
gkats, line 187: :wrench: I think we will always have `PrivateMetadata` here, it's ok not to `omitempty`.

clerk/organizations.go
```diff
@@ -161,3 +161,60 @@ func (s *OrganizationsService) CreateInvitation(params CreateOrganizationInvitat
...
+	Role            string          `json:"role"`
+	CreatedAt       int64           `json:"created_at"`
+	UpdatedAt       int64           `json:"updated_at"`
+
+	Organization   *Organization               `json:"organization"`
+	PublicUserData *organizationPublicUserData `json:"public_user_data,omitempty"`
+}
+
+type OrganizationMembershipResponse struct {
```
gkats, line 196:
```suggestion
type ListOrganizationMembershipsResponse struct {
```

clerk/organizations.go
```diff
@@ -161,3 +161,60 @@ func (s *OrganizationsService) CreateInvitation(params CreateOrganizationInvitat
...
+	Data       []OrganizationMembership `json:"data"`
+	TotalCount int64                    `json:"total_count"`
+}
+
+func (s *OrganizationsService) ListMemberships(params ListMembershipsParams) (*OrganizationMembershipResponse, error) {
```
gkats, line 201:
```suggestion
func (s *OrganizationsService) ListOrganizationMemberships(params ListMembershipsParams) (*ListOrganizationMembershipsResponse, error) {
```

gkats (changes requested):
:+1: Looks good overall, I think we can further improve on a few thinks.

AlexNti:
Thanks for the comments @gkats, i have update the code. Please check again

clerk/organizations.go
```diff
@@ -161,3 +161,56 @@ func (s *OrganizationsService) CreateInvitation(params CreateOrganizationInvitat
...
+	query := req.URL.Query()
+	if params.Limit != nil {
+		query.Set("limit", strconv.Itoa(*params.Limit))
+	}
+	if params.Offset != nil {
+		query.Set("offset", strconv.Itoa(*params.Offset))
+	}
+	req.URL.RawQuery = query.Encode()
+
+	var membershipsResponse *OrganizationMembershipResponse
```
gkats, line 210:
```suggestion
	var membershipsResponse *ListOrganizationMembershipsResponse
```

gkats (approved):

PR #95, opened 2023-01-24
Add Create and Update membership functions.

AlexNti:
## Type of change

Introduce two new functions that will allow us to update and create memberships.

### DIFF
--- a/clerk/organizations.go
+++ b/clerk/organizations.go
@@ -214,3 +214,35 @@ func (s *OrganizationsService) ListMemberships(params ListOrganizationMembership
 	}
 	return membershipsResponse, nil
 }
+
+type CreateOrganizationMembershipParams struct {
+	UserID string `json:"user_id"`
+	Role   string `json:"role"`
+}
+
+func (s *OrganizationsService) CreateMembership(organizationID string, params CreateOrganizationMembershipParams) (*OrganizationMembership, error) {
+	req, _ := s.client.NewRequest(http.MethodPost, fmt.Sprintf("%s/%s/memberships", OrganizationsUrl, organizationID), &params)
+
+	var organizationMembership OrganizationMembership
+	_, err := s.client.Do(req, &organizationMembership)
+	if err != nil {
+		return nil, err
+	}
+	return &organizationMembership, nil
+}
+
+type UpdateOrganizationMembershipParams struct {
+	UserID string
+	Role   string `json:"role"`
+}
+
+func (s *OrganizationsService) UpdateMembership(organizationID string, params UpdateOrganizationMembershipParams) (*OrganizationMembership, error) {
+	req, _ := s.client.NewRequest(http.MethodPatch, fmt.Sprintf("%s/%s/memberships/%s", OrganizationsUrl, organizationID, params.UserID), &params)
+
+	var organizationMembership OrganizationMembership
+	_, err := s.client.Do(req, &organizationMembership)
+	if err != nil {
+		return nil, err
+	}
+	return &organizationMembership, nil
+}
--- a/tests/integration/organizations_test.go
+++ b/tests/integration/organizations_test.go
@@ -18,15 +18,15 @@ type organizationMetadata struct {
 func TestOrganizations(t *testing.T) {
 	client := createClient()
 
-	limit := 1
+	limit := 2
 	users, err := client.Users().ListAll(clerk.ListAllUsersParams{
 		Limit: &limit,
 	})
 	if err != nil {
 		t.Fatalf("Users.ListAll returned error: %v", err)
 	}
-	if len(users) != 1 {
-		t.Fatalf("Users.ListAll returned %d results, expected 1", len(users))
+	if len(users) != 2 {
+		t.Fatalf("Users.ListAll returned %d results, expected 2", len(users))
 	}
 
 	newOrganization, err := client.Organizations().Create(clerk.CreateOrganizationParams{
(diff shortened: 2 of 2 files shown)


clerk/organizations.go
```diff
@@ -214,3 +214,37 @@ func (s *OrganizationsService) ListMemberships(params ListOrganizationMembership
...
+
+	var organizationMembership OrganizationMembership
+	_, err := s.client.Do(req, &organizationMembership)
+	if err != nil {
+		return nil, err
+	}
+	return &organizationMembership, nil
+}
+
+type UpdateOrganizationMembership struct {
```
gkats, line 235:
```suggestion
type UpdateOrganizationMembershipParams struct {
```

clerk/organizations.go
```diff
@@ -214,3 +214,37 @@ func (s *OrganizationsService) ListMemberships(params ListOrganizationMembership
 	}
 	return membershipsResponse, nil
 }
+
+type CreateOrganizationMembershipParams struct {
+	OrganizationID string
+	UserID         string `json:"user_id" form:"user_id"`
+	Role           string `json:"role" form:"role"`
```
gkats, line 221:
```suggestion
	Role           string `json:"role"`
```

clerk/organizations.go
```diff
@@ -214,3 +214,37 @@ func (s *OrganizationsService) ListMemberships(params ListOrganizationMembership
...
+}
+
+func (s *OrganizationsService) CreateMembership(params CreateOrganizationMembershipParams) (*OrganizationMembership, error) {
```
gkats, line 224: :wrench: Let's change the method signature so that it accepts the organization ID as a parameter. We do the same for the rest of the organization methods.

```go
type CreateOrganizationMembershipParams struct {
  UserID string `json:"user_id"`
  Role string  `json:"role"`
}

func (s *OrganizationService) CreateMembership(organizationID string, params CreateOrganizationMembershipParams) (*OrganizationMembership, error) {}
```

clerk/organizations.go
```diff
@@ -214,3 +214,37 @@ func (s *OrganizationsService) ListMemberships(params ListOrganizationMembership
...
+	return &organizationMembership, nil
+}
+
+type UpdateOrganizationMembership struct {
+	OrganizationID string
+	UserID         string
+	Role           string `json:"role"`
+}
+
+func (s *OrganizationsService) UpdateMembership(params UpdateOrganizationMembership) (*OrganizationMembership, error) {
```
gkats, line 241: :wrench: Let's change the signature here as well to be in line with the rest of our methods. I don't think it's necessarily better, but at least it follows a common pattern. :/ 

```go
type UpdateOrganizationMembershipParams struct {
  Role string `json:"role"`
}

func (s *OrganizationsService) UpdateMembership(organizationID, userID string, params UpdateOrganizationMembershipParams) (*OrganizationMembership, error) {}
```

gkats (changes requested):
Left some comments for a few changes in order to follow the same pattern as the rest of the methods in this file.

clerk/organizations.go
AlexNti: @gkats I got as reference the `CreateInvitation()` which was the last one added on the organizations, in that function the organizationId is being retrieved from `params`. 
Shall i change this one as well?

gkats: :disappointed: Thanks for the reference. Looks like we're all over the place. 

Let's ask for more opinions on this one. cc @clerkinc/backend-team.

clerk/organizations.go
chanioxaris: Yes that's our convention. Pass organizationID as a separate parameter and then all the others if necessary with a struct

clerk/organizations.go
```diff
@@ -214,3 +214,37 @@ func (s *OrganizationsService) ListMemberships(params ListOrganizationMembership
...
+	return &organizationMembership, nil
+}
+
+type UpdateOrganizationMembershipParams struct {
+	OrganizationID string
+	UserID         string
+	Role           string `json:"role"`
+}
+
+func (s *OrganizationsService) UpdateMembership(params UpdateOrganizationMembershipParams) (*OrganizationMembership, error) {
```
chanioxaris, line 241: 🔧 Same for here, pass organizationID as a standalone parameter

clerk/organizations.go
```diff
@@ -214,3 +214,37 @@ func (s *OrganizationsService) ListMemberships(params ListOrganizationMembership
 	}
 	return membershipsResponse, nil
 }
+
+type CreateOrganizationMembershipParams struct {
+	OrganizationID string
+	UserID         string `json:"user_id" form:"user_id"`
```
chanioxaris, line 220: No need to use a form tag

```suggestion
	UserID         string `json:"user_id"`
```

chanioxaris (approved):

tests/integration/organizations_test.go
```diff
@@ -99,6 +99,30 @@ func TestOrganizations(t *testing.T) {
...
+		t.Fatalf("Organizations.CreateMembership returned error: %v", err)
+	}
+	if createdOrganizationMembership == nil {
+		t.Fatalf("Organizations.CreateMembership returned nil")
+	}
+	assert.Equal(t, createdOrganizationMembership.Role, "admin")
+
+	updatedOrganizationMembership, err := client.Organizations().UpdateMembership(newOrganization.ID, clerk.UpdateOrganizationMembershipParams{
+		UserID: organizationMemberships.Data[0].PublicUserData.UserID,
+		Role:   "guest_member",
```
gkats, line 116:
```suggestion
		Role:   "basic_member",
```
Let's change this to basic_member, since guest_member is not supported.

gkats (approved):

SokratisVidros (approved):

PR #96, opened 2023-01-25
chore: Populate ListMemberships function to accept params.

AlexNti:
## Type of change

- [X] 🌟 New feature (non-breaking change which adds functionality)

## Description

`ListMembership()` now can be used to filter the expected results,using query,email etc

### DIFF
--- a/clerk/organizations.go
+++ b/clerk/organizations.go
@@ -166,6 +166,13 @@ type ListOrganizationMembershipsParams struct {
 	OrganizationID string
 	Limit          *int
 	Offset         *int
+	UserIDs        []string `json:"user_id"`
+	EmailAddresses []string `json:"email_address"`
+	PhoneNumbers   []string `json:"phone_number"`
+	Usernames      []string `json:"username"`
+	Web3Wallets    []string `json:"web3_wallet"`
+	OrderBy        *string  `json:"order_by"`
+	Query          *string  `json:"query"`
 }
 
 type ΟrganizationMembershipPublicUserData struct {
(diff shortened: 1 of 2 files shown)


clerk/organizations.go
```diff
@@ -166,6 +166,13 @@ type ListOrganizationMembershipsParams struct {
 	OrganizationID string
 	Limit          *int
 	Offset         *int
+	UserIDs        []string
+	EmailAddresses []string
+	PhoneNumbers   []string
+	Usernames      []string
+	Web3Wallets    []string
+	OrderBy        string
+	Query          string
```
chanioxaris, line 175: 🔧 I believe these (`OrderBy` & `Query`) are optional so the type should be `*string` instead

clerk/organizations.go
```diff
@@ -198,6 +205,16 @@ func (s *OrganizationsService) ListMemberships(params ListOrganizationMembership
 	endpoint := fmt.Sprintf("%s/%s/memberships", OrganizationsUrl, params.OrganizationID)
 	req, _ := s.client.NewRequest(http.MethodGet, endpoint)
 
+	s.client.Users().addUserSearchParamsToRequest(req, ListAllUsersParams{
```
chanioxaris, line 208: Not feeling that confident to reuse this as it is, as it's user specific. I would prefer to refactor this helper a bit to be generic. Then we can also use it for this operation as well

AlexNti: I had the same feeling, but i said to give it a go! I will create a new function inside organizations.

AlexNti: I had the same feeling for this, do you think it will be better to create a new function for the organizations? I was thinking also to put a new function in a shared file so both of them have access. Do we have this option here?

chanioxaris (approved):

PR #97, opened 2023-01-27
chore: Add DeleteMembership function

AlexNti:
## Type of change

## Description

We introduce  DeleteMembership function, that will allow us to delete a membership from an organization using the organizationid and the userId.


clerk/organizations.go
```diff
@@ -296,3 +296,18 @@ func (s *OrganizationsService) UpdateMembership(organizationID string, params Up
 	}
 	return &organizationMembership, nil
 }
+
+type DeleteOrganizationMembershipParams struct {
+	UserID string
+}
+
+func (s *OrganizationsService) DeleteMembership(organizationID string, params DeleteOrganizationMembershipParams) (*OrganizationMembership, error) {
```
chanioxaris, line 304: 🔧 Instead of passing a struct here, we must always accept userID as a parameter
```suggestion
func (s *OrganizationsService) DeleteMembership(organizationID, userID string) (*OrganizationMembership, error) {
```

AlexNti: @chanioxaris  Code has been update! 💯

chanioxaris (approved):

PR #98, opened 2023-01-31
chore: Add role search param for meberships

AlexNti:

 Add role as search param to filter the memberships list

### DIFF
--- a/clerk/organizations.go
+++ b/clerk/organizations.go
@@ -166,6 +166,7 @@ type ListOrganizationMembershipsParams struct {
 	OrganizationID string
 	Limit          *int
 	Offset         *int
+	Roles          []string `json:"role"`
 	UserIDs        []string `json:"user_id"`
 	EmailAddresses []string `json:"email_address"`
 	PhoneNumbers   []string `json:"phone_number"`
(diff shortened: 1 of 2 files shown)


tests/integration/organizations_test.go
```diff
@@ -143,7 +143,7 @@ func TestOrganizations(t *testing.T) {
 	assert.Equal(t, len(organizationMembershipsWithQuery.Data), 1)
 	assert.Equal(t, organizationMembershipsWithQuery.TotalCount, int64(1))
 
-	// Should return 0 results when using the email as search param and email does not exist
+	// Should return 0 results when using the email as search param and email does not exists
```
gkats, line 146: :wrench: The comment was correct. :smile:

clerk/organizations.go
```diff
@@ -228,6 +229,11 @@ func (s *OrganizationsService) addMembersSearchParamsToRequest(r *http.Request,
 			query.Add("user_id", userID)
 		}
 	}
+	if params.Roles != nil {
+		for _, role := range params.Roles {
+			query.Add("user_id", role)
```
gkats, line 234: :question: Is `"user_id"` intentional here? Shouldn't the parameter have to do with role?

AlexNti: This was a mistake. i have replace it with role.

clerk/organizations.go
```diff
@@ -228,6 +229,11 @@ func (s *OrganizationsService) addMembersSearchParamsToRequest(r *http.Request,
 			query.Add("user_id", userID)
 		}
 	}
+	if params.Roles != nil {
```
gkats, line 232: :upside_down_face: No need to check for `nil` here. If the slice is `nil` the range won't iterate at all. 

https://go.dev/play/p/-yM16GQvnOY

gkats (approved):

chanioxaris (approved):

PR #101, opened 2023-02-10
chore: Deprecate cookieless_dev setting

yourtallness:
## Type of change

## Description

Now that dashboard has switched to using the `url_based_session_syncing` setting, marking `cookieless_dev` as deprecated.

### DIFF
--- a/clerk/instances.go
+++ b/clerk/instances.go
@@ -37,6 +37,8 @@ type UpdateInstanceParams struct {
 	// CookielessDev can be used to enable the new mode in which no third-party
 	// cookies are used in development instances. Make sure to also enable the
 	// setting in Clerk.js
+	//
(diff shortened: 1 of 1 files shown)


gkats (approved):

chanioxaris (changes requested):
We must not remove the field as it will be a breaking change. We should mark it as deprecated instead

yourtallness:
> We must not remove the field as it will be a breaking change. We should mark it as deprecated instead

Will check if anyone other than our dashboard is using it.

yourtallness:
> We must not remove the field as it will be a breaking change. We should mark it as deprecated instead

@chanioxaris Changed to a deprecation warning

chanioxaris (approved):

PR #103, opened 2023-03-01
fix: interstitial of failed token verification

dimkl:
## Type of change

## Description

Set signed-out as default request state and return interstitial ONLY when the token verification fails with expired or invalid_iat errors.

Extra actions:
- [x] manual tests

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,1 @@
+* @clerkinc/backend-team
--- a/clerk/middleware_v2.go
+++ b/clerk/middleware_v2.go
@@ -2,12 +2,15 @@ package clerk
 
 import (
 	"context"
+	"errors"
 	"net"
 	"net/http"
 	"net/url"
 	"regexp"
 	"strconv"
 	"strings"
+
+	"gopkg.in/square/go-jose.v2/jwt"
 )
 
 var urlSchemeRe = regexp.MustCompile(`(^\w+:|^)\/\/`)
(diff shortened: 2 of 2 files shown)


clerk/middleware_v2.go
```diff
@@ -130,13 +133,26 @@ func WithSessionV2(client Client, verifyTokenOptions ...VerifyTokenOption) func(
...
-				next.ServeHTTP(w, r.WithContext(ctx))
+
+			if err == nil {
+				if claims.IssuedAt != nil && clientUatTs <= int64(*claims.IssuedAt) {
+					ctx := context.WithValue(r.Context(), ActiveSessionClaims, claims)
+					next.ServeHTTP(w, r.WithContext(ctx))
+					return
+				}
+
+				renderInterstitial(client, w)
```
chanioxaris, line 144: ❓ Is there is no error during token verification, why are we still rendering the interstitial?

dimkl: Because of this condition in Interstitial flow diagram:
Screenshot 2023-03-01 at 16 48 37

CODEOWNERS
```diff
@@ -0,0 +1 @@
+* @clerkinc/backend-team
```
dimkl, line 1: @SokratisVidros could you also give ` @clerkinc/backend-team` team write access for this repo (just like you did in clerk-sdk-ruby).

SokratisVidros: Done

clerk/middleware_v2.go
dimkl: @chanioxaris does this makes any sense ?

chanioxaris: Yes, thanks for the explanation

chanioxaris (approved):

PR #111, opened 2023-03-23
fix: Change UpdateUserMetadata to proper type

dgocoder:
## Type of change

## Description

Fixes issue with incorrect type on UpdateUserMetadata which is currently set to `interface{}` to `json.RawMessage` which is what is needed to make the request work however anything currently passed into it such as a map[string]string is currently valid.

### Related Issue (optional)

https://github.com/clerkinc/clerk-sdk-go/issues/109

### DIFF
--- a/clerk/users.go
+++ b/clerk/users.go
@@ -241,9 +241,9 @@ func (s *UsersService) Update(userId string, updateRequest *UpdateUser) (*User, 
 }
 
 type UpdateUserMetadata struct {
-	PublicMetadata  interface{} `json:"public_metadata"`
-	PrivateMetadata interface{} `json:"private_metadata"`
-	UnsafeMetadata  interface{} `json:"unsafe_metadata"`
+	PublicMetadata  json.RawMessage `json:"public_metadata"`
+	PrivateMetadata json.RawMessage `json:"private_metadata"`
+	UnsafeMetadata  json.RawMessage `json:"unsafe_metadata"`
 }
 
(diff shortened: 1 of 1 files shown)


SokratisVidros (approved):

gkats:
@dgocoder Thank you so much for the PR. :heart: 

How would you feel about changing the [metadata fields](https://github.com/clerkinc/clerk-sdk-go/blob/1fef0da955cf5ed9b7feb7fd1e474e307db57b13/clerk/users.go#L222) in the `UpdateUser` struct so that they are typed as `*json.RawMessage`? Similar to how it is for [creating a user](https://github.com/clerkinc/clerk-sdk-go/blob/1fef0da955cf5ed9b7feb7fd1e474e307db57b13/clerk/users.go#L67).

dgocoder:
@gkats  No issue in doing so however it seems across the board there is discrepancies. For instance in Org it is not a pointer and instead has omit empty on the json value. 

https://github.com/clerkinc/clerk-sdk-go/blob/efd909d8d336ce7af469660779bf5f38dc706581/clerk/organizations.go#L32

I'm happy to change it but I followed pattern suggested in the issue which you referenced the org which did not use a pointer.

gkats:
> however it seems across the board there are discrepancies

@dgocoder You're right, apologies for this. Let's not increase the scope of this PR. We'll tackle all discrepancies one at a time.

:raised_hands: Thanks again for your contribution!

PR #116, opened 2023-04-10
chore: Add UpdateHomeURL function

AlexNti:

We add UpdateHomeURL function that will allow us to update the domain url of the instance.

### DIFF
--- a/clerk/instances.go
+++ b/clerk/instances.go
@@ -106,3 +106,17 @@ func (s *InstanceService) UpdateOrganizationSettings(params UpdateOrganizationSe
 	}
 	return &organizationSettingsResponse, nil
 }
+
+type UpdateHomeURLParams struct {
+	HomeURL string `json:"home_url"`
+}
+
+func (s *InstanceService) UpdateHomeURL(params UpdateHomeURLParams) error {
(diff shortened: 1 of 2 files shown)


clerk/instances.go
```diff
@@ -106,3 +106,18 @@ func (s *InstanceService) UpdateOrganizationSettings(params UpdateOrganizationSe
...
+
+type UpdateHomeURLParams struct {
+	HomeURL string `json:"home_url"`
+}
+
+func (s *InstanceService) UpdateHomeURL(params UpdateHomeURLParams) error {
+	req, _ := s.client.NewRequest(http.MethodPost, "instance/change_domain", &params)
+
+	var response interface{}
+	_, err := s.client.Do(req, &response)
```
chanioxaris, line 118: 🔧 The endpoint doesn't return a response so we should change the code as follow. Please make sure that still will work
```suggestion
	_, err := s.client.Do(req, nil)
```

chanioxaris (approved):

clerk/instances.go
AlexNti: Replace it and still working.

chanioxaris: Awesome 💯

PR #119, opened 2023-04-25
chore: Add support of order by at organizations

AlexNti:

We Introduce a new parameter order_by that will allow us to order the organizations and organization memberships

### DIFF
--- a/tests/integration/organizations_test.go
+++ b/tests/integration/organizations_test.go
@@ -103,6 +103,22 @@ func TestOrganizations(t *testing.T) {
 	assert.NotEmpty(t, newOrganization2.ID)
 	assert.Equal(t, "my-org-2", newOrganization2.Name)
 
+	// Should short  Organizations by name DESC
+	orderBy := "-name"
+	organizations, err = client.Organizations().ListAll(clerk.ListAllOrganizationsParams{
+		OrderBy: &orderBy,
+	})
+	assert.NoError(t, err)
+	assert.Equal(t, organizations.Data[0].ID, newOrganization2.ID)
+
+	orderBy = "+name"
+	// Should short  Organizations by name ASC
+	organizations, err = client.Organizations().ListAll(clerk.ListAllOrganizationsParams{
+		OrderBy: &orderBy,
+	})
+	assert.NoError(t, err)
+	assert.Equal(t, organizations.Data[0].ID, newOrganization.ID)
+
 	// Should return 1 organization - exact match on ID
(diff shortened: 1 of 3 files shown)


tests/integration/organizations_test.go
```diff
@@ -103,6 +103,27 @@ func TestOrganizations(t *testing.T) {
...
+	})
+
+	assert.Equal(t, organizations.Data[0].ID, newOrganization.ID)
+
+	if err != nil {
+		t.Fatalf("Organizations.ListAll order by name returned error: %v", err)
+	}
+	if organizations == nil {
+		t.Fatalf("Organizations.ListAll order by name returned nil")
+	}
```
dimkl, line 125: 🙃 Should use use assertions instead of `if`s?
```
assert.Nil(t, err, "Organizations.ListAll order by name returned error");
assert.NotNil(t, organization, "Organizations.ListAll order by name returned nil");
```

dimkl (approved):

tests/integration/organizations_test.go
gkats: There's also 
```go
assert.NoError(t, err)
```
for errors.

gkats (approved):

tests/integration/organizations_test.go
```diff
@@ -103,6 +103,22 @@ func TestOrganizations(t *testing.T) {
 	assert.NotEmpty(t, newOrganization2.ID)
 	assert.Equal(t, "my-org-2", newOrganization2.Name)
 
+	// Should short  Organizations by name DESC
+	orderBy := "-name"
+	organizations, err = client.Organizations().ListAll(clerk.ListAllOrganizationsParams{
+		OrderBy: &orderBy,
```
chanioxaris, line 109: 🔧 In a recent commit i introduced some helpers to convert string to *string for example. Let's use those to make our life easier

```suggestion
		OrderBy: stringToPtr("-name"),
```

tests/integration/organizations_test.go
```diff
@@ -103,6 +103,22 @@ func TestOrganizations(t *testing.T) {
...
+	})
+	assert.NoError(t, err)
+	assert.Equal(t, organizations.Data[0].ID, newOrganization2.ID)
+
+	orderBy = "+name"
+	// Should short  Organizations by name ASC
+	organizations, err = client.Organizations().ListAll(clerk.ListAllOrganizationsParams{
+		OrderBy: &orderBy,
```
chanioxaris, line 117:
```suggestion
		OrderBy: stringToPtr("+name"),
```

chanioxaris (approved):

Issue #125, opened 2023-05-09
Add filtering to Sessions().ListAll()


danthegoodman1:
https://clerk.com/docs/reference/backend-api/tag/Sessions#operation/GetSessionList docs have filtering options, but they are not accessible from the golang SDK

petergeorgas:
I would love to do this! I can work on it in a bit.

petergeorgas:
This can be closed

PR #127, opened 2023-05-11
Add filtering to Sessions().ListAll()

petergeorgas:
## Type of change

## Description

Add Filtering to `Sessions().ListAll()` to match the API. 

### Related Issue (optional)

#125

### DIFF
--- a/clerk/sessions.go
+++ b/clerk/sessions.go
@@ -35,7 +35,19 @@ const (
 	STATUS_REVOKED   SessionStatus = "revoked"
 )
 
-func (s *SessionsService) ListAll(params ListAllSessionsParams) ([]Session, error) {
+func (s *SessionsService) ListAll() ([]Session, error) {
+	sessionsUrl := "sessions"
+	req, _ := s.client.NewRequest("GET", sessionsUrl)
+
+	var sessions []Session
+	_, err := s.client.Do(req, &sessions)
+	if err != nil {
+		return nil, err
+	}
+	return sessions, nil
+}
+
+func (s *SessionsService) ListAllWithFiltering(params ListAllSessionsParams) ([]Session, error) {
 	sessionsUrl := "sessions"
 	req, _ := s.client.NewRequest("GET", sessionsUrl)
 
--- a/clerk/sessions.go
+++ b/clerk/sessions.go
@@ -26,13 +26,13 @@ type ListAllSessionsParams struct {
 type SessionStatus string
 
 const (
-	STATUS_ABANDONED SessionStatus = "abandoned"
-	STATUS_ACTIVE    SessionStatus = "active"
-	STATUS_ENDED     SessionStatus = "ended"
-	STATUS_EXPIRED   SessionStatus = "expired"
-	STATUS_REMOVED   SessionStatus = "removed"
-	STATUS_REPLACED  SessionStatus = "replaced"
-	STATUS_REVOKED   SessionStatus = "revoked"
+	SessionStatusAbandoned SessionStatus = "abandoned"
+	SessionStatusActive    SessionStatus = "active"
+	SessionStatusEnded     SessionStatus = "ended"
+	SessionStatusExpired   SessionStatus = "expired"
+	SessionStatusRemoved   SessionStatus = "removed"
+	SessionStatusReplaced  SessionStatus = "replaced"
+	SessionStatusRevoked   SessionStatus = "revoked"
 )
 
 func (s *SessionsService) ListAll() ([]Session, error) {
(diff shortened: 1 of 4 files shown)


clerk/sessions.go
```diff
@@ -15,10 +15,47 @@ type Session struct {
...
+	STATUS_ABANDONED SessionStatus = "abandoned"
+	STATUS_ACTIVE    SessionStatus = "active"
+	STATUS_ENDED     SessionStatus = "ended"
+	STATUS_EXPIRED   SessionStatus = "expired"
+	STATUS_REMOVED   SessionStatus = "removed"
+	STATUS_REPLACED  SessionStatus = "replaced"
+	STATUS_REVOKED   SessionStatus = "revoked"
+)
+
+func (s *SessionsService) ListAll(params ListAllSessionsParams) ([]Session, error) {
```
dimkl, line 38: 🔧 i would expect the params to be optional to avoid making this a BREAKING CHANGE.
```
ListAll(params *ListAllSessionsParams) 
```

clerk/sessions_test.go
```diff
@@ -23,7 +23,7 @@ func TestSessionsService_ListAll_happyPath(t *testing.T) {
 	var want []Session
 	_ = json.Unmarshal([]byte(expectedResponse), &want)
 
-	got, _ := client.Sessions().ListAll()
+	got, _ := client.Sessions().ListAll(ListAllSessionsParams{})
```
dimkl, line 26: 🔧 revert these changes to test that current sdk usage does not break.

clerk/sessions_test.go
```diff
@@ -8,7 +8,7 @@ import (
 	"testing"
 )
 
-func TestSessionsService_ListAll_happyPath(t *testing.T) {
+func TestSessionsService_ListAll_happyPath_noParams(t *testing.T) {
```
dimkl, line 11: 🔧 Could you please add 2 tests:
- `TestSessionsService_ListAll_pagination_and_filtering_params`
- `TestSessionsService_ListAll_pagination_and_filtering_empty_params`

test to assert that the added functionality works as expected?

clerk/sessions.go
gkats: As @dimkl mentioned, simply adding the params here would be a breaking change. We're not ready for a major version bump yet unfortunately.

Even adding them as a pointer would be a breaking change, so we can't do `ListAll(p *ListAllSessionsParams)`.

I think the correct way to approach this would be to add a separate method on the SessionsService. Something like

```go
func (s *SessionService) ListAllWithFiltering(params ListAllWithFilteringSessionsParams) {}
```

@petergeorgas What do you think about that? Open to any other suggestion of course. The important thing is to be backwards compatible.

petergeorgas: Good point! Just made the changes!

clerk/sessions.go
```diff
@@ -15,6 +15,26 @@ type Session struct {
...
+
+const (
+	STATUS_ABANDONED SessionStatus = "abandoned"
+	STATUS_ACTIVE    SessionStatus = "active"
+	STATUS_ENDED     SessionStatus = "ended"
+	STATUS_EXPIRED   SessionStatus = "expired"
+	STATUS_REMOVED   SessionStatus = "removed"
+	STATUS_REPLACED  SessionStatus = "replaced"
+	STATUS_REVOKED   SessionStatus = "revoked"
+)
```
gkats, line 36: :wrench: Consider prefixing these constants with `SESSION_`. Since everything's inside the Clerk package, if we ever want to add similar constants to other resources, there's going to be a naming collision.

Also, let's write them in camel case, maybe it's more Go idiomatic.

```go
// This is for sessions
const SessionStatusAbandoned SessionStatus = "abandoned"

// This is for some other resource
const OtherResourceStatusAbandoned = "abandoned"

// They are both exported by the clerk package
clerk.OtherResourceStatusAbandoned
clerk.SessionStatusAbandoned
```

gkats (changes requested):
Awesome work @petergeorgas!

Left one more comment regarding the session status constants. Let's update this and then we're good to :shipit: 

Thank you so much for taking care of this!

clerk/sessions.go
petergeorgas: Sounds good! I never know what to do with constants... thanks!

gkats (approved):

gkats:
@petergeorgas Thank you so much for the contribution!! :heart_eyes: :tada:

petergeorgas:
> @petergeorgas Thank you so much for the contribution!! 😍 🎉

Happy to help! Looking forward to tackling other issues in the future.

PR #140, opened 2023-06-22
feat: Update organization logo

gkats:
## Type of change

## Description

Added a new method on the organizations service that allows to set or update an organization's logo. 

Accepts the organization ID, the file for the logo and the uploader user ID as parameters. Triggers a multipart/form-data request to our Backend API to update the organization logo.


clerk/organizations.go
```diff
@@ -66,6 +69,44 @@ func (s *OrganizationsService) Update(organizationID string, params UpdateOrgani
 	return &organization, nil
 }
 
+type UpdateOrganizationLogoParams struct {
+	File           multipart.File
+	Filename       string
```
gkats, line 74: I couldn't find an elegant way to get the filename from the file. If anybody has any suggestion, please help. :smile:

gkats: Made the parameter optional.

tests/integration/organizations_test.go
```diff
@@ -86,6 +88,21 @@ func TestOrganizations(t *testing.T) {
 	}
 	assert.Equal(t, slug, *updatedOrganization.Slug)
 
+	// Update organization logo
+	assert.Nil(t, updatedOrganization.LogoURL)
+	filename := "200x200-grayscale.jpg"
+	file, err := os.Open(fmt.Sprintf("../../testdata/%s", filename))
```
georgepsarakis, line 94: [Minor] You can use [path.Join](https://pkg.go.dev/path#Join) to form the path.

clerk/organizations_test.go
```diff
@@ -199,6 +202,80 @@ func TestOrganizationsService_ListAll_invalidServer(t *testing.T) {
...
+			if header.Filename != filename {
+				t.Errorf("expected %s, got %s", filename, header.Filename)
+			}
+			defer fileParam.Close()
+
+			var (
+				got  []byte
+				want []byte
+			)
+			_, err = fileParam.Read(got)
```
georgepsarakis, line 244: 🔧 I may be missing something here but the `Read` methods read number of bytes equal to the length of the slice that's passed to them. If we pass an empty slice then nothing is read I think.

clerk/organizations_test.go
```diff
@@ -199,6 +202,80 @@ func TestOrganizationsService_ListAll_invalidServer(t *testing.T) {
...
+			if err != nil {
+				t.Fatal(err)
+			}
+			_, err = file.Read(want)
```
georgepsarakis, line 248: ❓ Isn't the file handle already seeked to its end within the `UpdateLogo` function? I think you may need to seek to the start or re-open? 🤔

clerk/organizations_test.go
gkats: Great catch, thanks! Updated the code to set the slice length to each file's size before reading.

clerk/organizations_test.go
gkats: Updated to move reset the seeker.

tests/integration/organizations_test.go
gkats: :+1: Updated here and the `clerk/organizations_test.go` file.

georgepsarakis (approved):

PR #142, opened 2023-06-27
feat: Delete organization logo

gkats:
## Type of change

## Description

Added a new DeleteLogo() method on the organizations service. 

Accepts the organization ID as a parameter and removes the logo from the organization.

### DIFF
--- a/clerk/organizations_test.go
+++ b/clerk/organizations_test.go
@@ -286,6 +286,27 @@ func TestOrganizationsService_UpdateLogo(t *testing.T) {
 	}
 }
 
+func TestOrganizationsService_DeleteLogo(t *testing.T) {
+	client, mux, _, teardown := setup("token")
+	defer teardown()
+
+	organizationID := "org_123"
+	mux.HandleFunc(
+		fmt.Sprintf("/organizations/%s/logo", organizationID),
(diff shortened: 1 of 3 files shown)


clerk/organizations_test.go
```diff
@@ -286,6 +286,27 @@ func TestOrganizationsService_UpdateLogo(t *testing.T) {
...
+	client, mux, _, teardown := setup("token")
+	defer teardown()
+
+	organizationID := "org_123"
+	mux.HandleFunc(
+		fmt.Sprintf("/organizations/%s/logo", organizationID),
+		func(w http.ResponseWriter, req *http.Request) {
+			testHttpMethod(t, req, http.MethodDelete)
+			testHeader(t, req, "Authorization", "Bearer token")
+			fmt.Fprint(w, `{"id":"img_1234"}`)
```
georgepsarakis, line 299: ❓ Doesn't the response contain the organization? That's only for semantics though, but if that's the case, just to avoid confusion, consider `img_1234` -> `org_123` as id?

gkats: You're right, thanks for catching this. Updated.

georgepsarakis: 🚀

PR #149, opened 2023-08-02
fix: Introduce custom string to better handle unset attributes

fragoulis:
## Type of change

## Description

There are 2 separate issues we are trying to solve and both are needed in order to correctly encode and decode a data representation, to and from json, to and from go.

To solve decoding from json and correctly handle the different states a key/value pair can take, we introduce a custom String struct. The custom String can hold set/unset information, null/present information as well as the actual value if present.

The reason this is needed is because there is no other way to avoid the default behavior, which is to handle as same, a missing attribute and a present attribute set to null.

To solve encoding to json and again correctly represent the missing attributes and also support presence as null, we cannot blindly re-purpose the UpdateDomainParams struct which we use for the incoming request.

We use a custom struct that utilizes omitempty to make sure we do not set attributes for the bapi payload that were not set in the incoming dapi payload.

### Related Issue (optional)

Fix USR-261

### DIFF
--- a/clerk/domains.go
+++ b/clerk/domains.go
@@ -36,8 +36,8 @@ type CreateDomainParams struct {
 }
 
 type UpdateDomainParams struct {
-	Name     *string `json:"name,omitempty"`
-	ProxyURL *string `json:"proxy_url,omitempty"`
+	Name     String `json:"name"`
+	ProxyURL String `json:"proxy_url"`
 }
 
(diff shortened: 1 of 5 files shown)


chanioxaris (changes requested):
Indeed this is an issue that our SDK can't handle, but sadly this is a breaking change.
Let's add it to the list of SDK v2 tasks instead

PR #173, opened 2023-11-06
feat: Add organization roles endpoints

mzhong9723:
## Type of change

## Description

Add support for organization roles backend API endpoints


clerk/clerk.go
```diff
@@ -309,6 +313,10 @@ func (c *client) Organizations() *OrganizationsService {
 	return c.organizations
 }
 
+func (c *client) OrganizationRoles() *OrganizationRolesService {
```
chanioxaris, line 316: I am thinking that we probably need to differentiate between organization roles on the instance level and on the organization level. If in the future we are looking to add endpoints dedicated to the organization endpoints (e.g. /v1/organizations/:id/roles) we will have a name conflict here

clerk/organization_roles_test.go
```diff
@@ -0,0 +1,236 @@
...
+	_, err := client.OrganizationRoles().Delete(dummyOrgRoleID)
+	if err != nil {
+		t.Fatal(err)
+	}
+}
+
+const dummyOrgRoleID = "role_1mebQggrD3xO5JfuHk7clQ94ysA"
+
+const dummyOrgRoleJson = `{
+	"object": "organization_role",
```
chanioxaris, line 217:
```suggestion
	"object": "role",
```

clerk/organization_roles_test.go
```diff
@@ -0,0 +1,236 @@
...
+	"name": "custom role",
+	"key": "org:custom_role",
+	"description": "my org custom role",
+	"permissions": [],
+	"created_at": 1610783813,
+	"updated_at": 1610783813
+}`
+
+const dummyUpdateOrgRoleJson = `{
+	"object": "organization_role",
```
chanioxaris, line 228:
```suggestion
	"object": "role",
```

chanioxaris (approved):

clerk/organization_roles.go
```diff
@@ -0,0 +1,123 @@
...
+
+import (
+	"fmt"
+	"net/http"
+	"strconv"
+)
+
+type OrganizationRolesService service
+
+type OrganizationRole struct {
```
chanioxaris, line 11: Do we want to have separate types for each role context or make it a generic one `Role`? Whatever we think is the best, let's make sure to align the `Permission` struct name as well

clerk/clerk.go
gkats: I would also expect these methods to go into different services. 

```go
clerk.Instances().CreateOrganizationRole()
clerk.Organizations.CreateRole()
```
We could have one type for roles `type Role struct` but ideally the two methods above should accept different type of parameters.

clerk/organization_roles.go
mzhong9723: Let’s keep it separate for now - I think it’s easier to understand this way what the struct is for!

clerk/clerk.go
mzhong9723: Makes sense, I will refactor this! And yes I agree, let’s start with different types for now

PR #178, opened 2023-11-13
chore: Align naming for Roles and Permissions

chanioxaris:
## Type of change

## Description

In this commit we align the naming for the Role and Permission resources with the rest of our SDKs. We will use a more generic term of Role and Permission for every related endpoint

**The feature that uses these endpoints are under active development, thus these changes are not considered breaking**

### DIFF
--- a/clerk/instance_organization_permissions.go
+++ b/clerk/instance_organization_permissions.go
@@ -4,28 +4,12 @@ import (
 	"net/http"
 )
 
-type OrganizationPermission struct {
-	Object      string `json:"object"`
-	ID          string `json:"id"`
-	Name        string `json:"name"`
-	Key         string `json:"key"`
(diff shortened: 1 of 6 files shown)


clerk/instance_organization_permissions.go
```diff
@@ -4,7 +4,7 @@ import (
 	"net/http"
 )
 
-type OrganizationPermission struct {
+type Permission struct {
```
gkats, line 7: :question: Are we supposed to move these to different files as well?

gkats (approved):

clerk/instance_organization_permissions.go
chanioxaris: Sure thing, moved to separate files!

PR #183, opened 2023-11-16
feat: Add support for BAPI custom permissions endpoints

mzhong9723:
## Type of change

## Description

Add read, create, update, delete endpoints for custom permission

### DIFF
--- a/clerk/instance_organization_permissions.go
+++ b/clerk/instance_organization_permissions.go
@@ -1,6 +1,7 @@
 package clerk
 
 import (
+	"fmt"
 	"net/http"
 )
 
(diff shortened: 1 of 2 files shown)


clerk/instance_organization_permissions.go
```diff
@@ -24,3 +25,62 @@ func (s *InstanceService) ListOrganizationPermissions(params ListInstanceOrganiz
 	}
 	return response, nil
 }
+
+type CreateInstanceOrganizationPermissionParams struct {
```
gkats, line 29: :upside_down_face: Can't we name this `CreateOrganizationPermissionsParams`?

clerk/instance_organization_permissions.go
```diff
@@ -24,3 +25,62 @@ func (s *InstanceService) ListOrganizationPermissions(params ListInstanceOrganiz
...
+
+	var orgPermission Permission
+	_, err = s.client.Do(req, &orgPermission)
+	if err != nil {
+		return nil, err
+	}
+	return &orgPermission, nil
+}
+
+type UpdateInstanceOrganizationPermissionParams struct {
```
gkats, line 60: :upside_down_face: Same about the name, `UpdateOrganizationPermissionParams`.

And since it's the same parameters with Create, would it make sense to have only one type, `OrganizationPermissionParams`?

gkats (approved):

clerk/instance_organization_permissions.go
chanioxaris: What if we introduce the same method on the organization resource/service as well? As those are under the same package, there will be a name conflict

chanioxaris (approved):

clerk/instance_organization_permissions.go
mzhong9723: > What if we introduce the same method on the organization resource/service as well? As those are under the same package, there will be a name conflict

Yes let's keep it with `instance` in the name, at least for this iteration of the project.

clerk/instance_organization_permissions.go
mzhong9723: It is the same fields, except they are optional in the update params. Is there a better way to handle this given that?

Will merge this PR for now, but happy to open a new one to clean it up!

PR #195, opened 2023-12-08
chore: Update README to include Client constructor options

### LINKED ISSUE #194: Retries in http client
Transient errors are inevitable and therefore for most of our http clients we use a retrying http client (in our case `hashicorp/go-retryablehttp`, but that shouldn't matter).

But with Clerk we neither can provide such a client (as you're using an exact type for the http client instead of an interface), nor does your SDK itself allow to retry transient errors.

As you're only using a single method of the client, changing it to an interface should be a quick win

georgepsarakis:
Closes #194

## Type of change

## Description

- Add documentation and example for injecting a custom `*http.Client` instance for use by the SDK methods.

image

### Related Issue (optional)

Closes #194

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,6 +64,21 @@ follow the same structure that can be found in the [Backend API documentation](h
 
 For more examples on how to use the client, refer to the [examples](https://github.com/clerkinc/clerk-sdk-go/tree/main/examples/operations)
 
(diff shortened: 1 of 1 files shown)


chanioxaris (approved):
💯

PR #199, opened 2024-01-09
feat: Add helper methods to check for authorization using permissions and roles

mzhong9723:
- Add helper method to check permission/roles with session token
- add helper methods on `SessionClaims` struct
- upgrade to go 1.21 (using `slice.Contains()` which was introduced in 1.18, lmk if you think there's a reason we shouldn't update to 1.21)


.github/workflows/tests.yml
```diff
@@ -11,7 +11,7 @@ jobs:
   test:
     strategy:
       matrix:
-        go-version: [1.16.x]
+        go-version: [1.21.x]
```
georgepsarakis, line 14: ❓ Should we add all the supported versions in the matrix perhaps? Although Go doesn't break backwards compatibility there might be features implemented in 1.16+ that'll make the code not compile in older versions.

clerk/authorization.go
```diff
@@ -0,0 +1,47 @@
...
+func (c *client) CheckAuthorization(token string, params CheckAuthorizationParams) (bool, error) {
+	claims, err := c.VerifyToken(token)
+	if err != nil {
+		return false, err
+	}
+
+	permission := params.Permission
+	role := params.Role
+
+	if permission != "" && slices.Contains(claims.ActiveOrganizationPermissions, permission) {
```
georgepsarakis, line 23: 🔧 Continuing on my previous comment on the Go version support, any SDK user with a version older than 1.21 will have a problem upgrading since the `slices` package is not present in previous versions.

clerk/authorization.go
```diff
@@ -0,0 +1,47 @@
...
+type CheckAuthorizationParams struct {
+	Permission string
+	Role       string
+}
+
+// CheckAuthorization verifies if the user has the given permission or role.
+// Performing role checks is not considered a best-practice and
+// developers should avoid it as much as possible.
+// Usually, complex role checks can be refactored with a single permission check.
+func (c *client) CheckAuthorization(token string, params CheckAuthorizationParams) (bool, error) {
```
georgepsarakis, line 14: ❓ I'm not sure if the design has been already discussed and analyzed in depth but is the current API preferred over explicit methods such as:

```go
func (c *client) TokenHasPermission(token string, permission string) (bool, error) {}
func (c *client) TokenHasRole(token string, role string) (bool, error) {}
```

Perhaps we try to have parity with the JS SDK methods and I'm unaware of it. Asking this now, because it will be difficult to deprecate this helper afterwards!

clerk/authorization.go
```diff
@@ -0,0 +1,47 @@
...
+	if claims.ActiveOrganizationRole == role {
+		return true, nil
+	}
+
+	return false, nil
+}
+
+// CheckPermission checks if the user has the specific permission
+// in their session claims.
+func (s *SessionClaims) CheckPermission(permission string) bool {
```
georgepsarakis, line 36: ❓ Should these be defined at the file where the definition of the `SessionClaims` struct also resides?

clerk/authorization.go
```diff
@@ -0,0 +1,47 @@
...
+	}
+
+	permission := params.Permission
+	role := params.Role
+
+	if permission != "" && slices.Contains(claims.ActiveOrganizationPermissions, permission) {
+		return true, nil
+	}
+
+	if claims.ActiveOrganizationRole == role {
```
georgepsarakis, line 27: ❓ Isn't this equivalent to:

```suggestion
	if claims.CheckRole(role) {
```

clerk/authorization.go
georgepsarakis, line 23: ❓ Should we reuse the `SessionClaims.CheckPermission` method here?

```suggestion
	if permission != "" && claims.CheckPermission(permission) {
```

clerk/authorization_test.go
```diff
@@ -0,0 +1,90 @@
...
+	token, pubKey := testGenerateTokenJWT(t, dummySessionClaims, "kid")
+	client := c.(*client)
+	client.jwksCache.set(testBuildJWKS(t, pubKey, jose.RS256, "kid"))
+
+	// user has permission
+	hasPermission, err := c.CheckAuthorization(
+		token,
+		CheckAuthorizationParams{Permission: "org:billing:manage"},
+	)
+	if err != nil {
```
georgepsarakis, line 20: 🔧 We can use the `testify` package in tests:

```suggestion
	require.NoError(t, err)
```

clerk/authorization_test.go
```diff
@@ -0,0 +1,90 @@
...
+		t.Error(err)
+	}
+	if !hasPermission {
```
georgepsarakis, line 23: 🔧 Same as above, we can use `assert` here:

```go
assert.True(t, hasPermission, "Expected user to have permission: %s", "org:billing:manage")
```

go.mod
```diff
@@ -1,11 +1,17 @@
 module github.com/clerkinc/clerk-sdk-go
 
-go 1.16
+go 1.21
```
georgepsarakis, line 3: 🔧 I think a Go version upgrade may require a separate PR and (at least) a minor version bump, as this is a breaking change as far as I understand.

gkats (changes requested):
Let's avoid bumping the Go version in this PR.

Feels that it's a big change and we should make a conscious decision about it. We might (or might not) break anybody, since we rely on the promise of compatibility in 1.x versions.

However, it means that we'll be forcing some consumers to update. I don't think this is bad, let's just first decide that we want to do it.

By the way, I believe the minimum Go version still actively supported will soon be 1.20, so 1.16 is indeed old.

> Each major Go release is supported until there are two newer major releases. For example, Go 1.5 was supported until the Go 1.7 release, and Go 1.6 was supported until the Go 1.8 release. We fix critical problems, including [critical security problems](https://go.dev/security), in supported releases as needed by issuing minor revisions (for example, Go 1.6.1, Go 1.6.2, and so on).

clerk/authorization.go
chanioxaris: I agree with George on that, i believe the `CheckAuthorization()` helper doesn't make sense at this point. As a developer i would expect to check if a user has the proper role/permission using the sessionClaims, which are already populated and validated from the token and available on my Backend. I believe the methods on the sessionClaims are good enough to enforce elevated access control

mzhong9723:
Closing this PR in favor of new one that doesn’t include a go version update 🙂

PR #200, opened 2024-01-10
feat: Add session claims helper methods to check roles and permissions

mzhong9723:
Add `HasPermission()` and `HasRole()` to session claims

### DIFF
--- a/clerk/session_claims_test.go
+++ b/clerk/session_claims_test.go
@@ -1,0 +1,31 @@
+package clerk
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestSessionClaims_HasPermissiont(t *testing.T) {
+	// user has permission
+	hasPermission := dummySessionClaims.HasPermission("org:billing:manage")
+	assert.True(t, hasPermission)
+
+	// user has second permission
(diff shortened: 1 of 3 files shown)


agis (approved):

clerk/session_claims_test.go
```diff
@@ -0,0 +1,27 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestSessionClaims_HasPermissiont(t *testing.T) {
+	// user has permission
+	hasPermission := dummySessionClaims.HasPermission("org:billing:manage")
+	assert.True(t, hasPermission)
+
```
georgepsarakis, line 13: 🔧 You could also assert that the helper finds both permissions defined in the claims:

https://github.com/clerk/clerk-sdk-go/blob/fede4e5118f91993dce476862637c600a376a533/clerk/tokens_test.go#L56

This case may provide an additional guarantee that we perform an exhaustive search.

georgepsarakis (approved):

clerk/session_claims_test.go
mzhong9723: Will add! Thanks for the suggestion

PR #206, opened 2024-01-30
Groundwork for v2

gkats:
## Type of change

- [x] 🔨 Breaking change (fix or feature that would cause existing functionality)

## Description

:tada: Kicking off the v2 rewrite of our Go SDK. :tada: 

The work here is split into three commits. It might be easier to review commit by commit, because there's a lot of changed files.

- First commit deletes all existing files. 
- Second commit contains the basis for building API operations in our SDK.
- Third commit sets up a GH workflow for CI.

We're moving into a different direction with our SDK. This PR includes the core tools for implementing all API operations.

We have three basic interfaces in our SDK.

```go
// Backend send API requests to the Clerk API. The response will be 
// passed to the ResponseReader. 
type Backend interface {
  Call(context.Context, *APIRequest, ResponseReader)
}

// ResponseReader reads API responses.
type ResponseReader interface {
  Read(*APIResponse)
}

// Queryable can set query string parameters.
type Queryable interface {
  Add(url.Values)
}
```

Along with these interfaces we have some basic types that describe core API resources. These are
```go
// APIResource describes common functionality for all API resources. These could 
// be types for users, organizations, domains, etc.
// APIResource implements the ResponseReader interface. This way, we can build 
// resources out of API responses.
type APIResource struct {}

// APIParams describes common functionality for parameters that can be included in 
// API requests. 
type APIParams struct {}

// APIRequest is a wrapper over an http.Request for the Clerk API.
type APIRequest struct {}

// APIResponse is a wrapper over an http.Response for the Clerk API.
type APIResponse struct {}

// APIErrorResponse is a wrapper over an unsuccessful http.Response for the Clerk API.
// The Clerk API can return error responses, so the APIErrorResponse also implements the 
// Error interface.
type APIErrorResponse struct {}
```
A default Backend implementation is provided but not exported.

Using the above types, implementing an API operation should be easy. Take a look at the tests (clerk_test.go) for a few examples of how we would write API operations.

```go
package domain

func Create(ctx context.Context, params *clerk.DomainParams) (*clerk.Domain, error) {
  req := clerk.NewAPIRequest("POST", "/domains")
  req.SetBody(params)
  dmn := &clerk.Domain{}
  err := clerk.GetBackend().Call(ctx, req, dmn)
  return dmn, err
}
```

Finally, consuming the API operation that we implemented above would look like this
```go
clerk.SecretKey = "sk_live_123"
dmn, err := domain.Create(ctx, &clerk.DomainParams{})
```

As you might have guessed, the API and implementation is based on [Stripe's Go SDK](https://github.com/stripe/stripe-go/)


clerk.go
```diff
@@ -0,0 +1,360 @@
...
+	if err != nil {
+		return err
+	}
+
+	return b.do(req, apiReq.Params, setter)
+}
+
+func (b *defaultBackend) newRequest(ctx context.Context, apiReq *APIRequest) (*http.Request, error) {
+	path := b.URL + apiReq.Path
+	req, err := http.NewRequest(apiReq.Method, path, nil)
```
georgepsarakis, line 204: 🔧 I think you can pass the context here directly: https://pkg.go.dev/net/http#NewRequestWithContext

clerk.go
```diff
@@ -0,0 +1,360 @@
...
+	Backend Backend
+	mu      sync.RWMutex
+}
+
+// defaultHTTPTimeout is the default timeout on the http.Client used
+// by the library.
+const defaultHTTPTimeout = 5 * time.Second
+
+// The default HTTP client used for communication with the Clerk API.
+var httpClient = &http.Client{
```
georgepsarakis, line 302: ❓ Perhaps:

```suggestion
var defaultHTTPClient = &http.Client{
```

clerk_test.go
```diff
@@ -0,0 +1,344 @@
...
+
+// TestBackendCall_SuccessfulResponse_PostRequest tests that for POST
+// requests (or other mutating operations) we serialize all parameters
+// in the request body.
+func TestBackendCall_SuccessfulResponse_PostRequest(t *testing.T) {
+	ctx := context.Background()
+	name := "the-name"
+	rawJSON := `{"id":"res_123","object":"resource"}`
+
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
georgepsarakis, line 181: ❓ Would a package like [httpmock](https://github.com/jarcoal/httpmock) help us in some way?

clerk.go
```diff
@@ -0,0 +1,360 @@
...
+	if err != nil || apiError.Errors == nil {
+		// This is probably not an expected API error.
+		// Return the raw server response.
+		return errors.New(string(body))
+	}
+	return apiError
+}
+
+// The active Backend
+var backend api
```
georgepsarakis, line 288: ❓ I'm not sure I understand the use of these globals, along with the `Get/SetBackend` functions. I would maybe expect that a `NewAPIClient` method (if I understand the parity with the `Backend` interface here), returns a fully isolated client instance. Any chance we need to simplify some aspects here?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,45 @@
+name: CI
+
+on:
+  pull_request:
+  push:
+    branches:
+      - $default-branch
```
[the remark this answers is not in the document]
gkats: I think so. Copied it over from our current github workflow file, assuming it resolves to `main`. 

https://github.com/clerk/clerk-sdk-go/blob/b8f5d47d1ee62ac734a9c61d56dfdf6a33d103c3/.github/workflows/tests.yml#L4

clerk.go
```diff
@@ -0,0 +1,360 @@
...
+	clerkVersion string = "v1"
+)
+
+const (
+	// APIURL is the base URL for the Clerk API.
+	APIURL string = "https://api.clerk.com"
+)
+
+// SecretKey is the Clerk secret key. Configured on a package level.
+var SecretKey string
```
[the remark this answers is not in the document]
gkats: @georgepsarakis Can you think of a problem with exporting the `SecretKey` on the package level?

The benefit is that you can set your key with a simple API.

```go
clerk.SecretKey = "sk_live_123"
```

clerk.go
```diff
@@ -0,0 +1,360 @@
...
+func (resp *APIResponse) Success() bool {
+	return resp.StatusCode < 400
+}
+
+// NewAPIResponse creates an APIResponse from the passed http.Response
+// and the raw response body.
+func NewAPIResponse(resp *http.Response, body json.RawMessage) *APIResponse {
+	return &APIResponse{
+		Header:     resp.Header,
+		TraceID:    resp.Header.Get("Clerk-Trace-Id"),
```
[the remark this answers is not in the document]
gkats: Sure, but I'd rather wait until there's demand or a certain use case. The point is to hide the details and provide a dedicated field for the trace ID.

If we ever decide to move away from headers, we won't have to change anything.

clerk_test.go
gkats: Thanks for the suggestion, but I don't see any real benefit to adding this dependency. 

Unless you have a specific scenario in mind, or you feel that the tests are hard to follow.

To give you some context, I wasn't even sure if the dependency to `testify` was worth it. :smile:

clerk.go
gkats: It depends. 

Having a "singleton" global API client allows for two things:

1. Simple API for operations
```go
clerk.SecretKey = "sk_live_123"
dmn, err := domain.Create(ctx, &DomainParams{})
```

2. Ability to stub out the client for tests
```go
clerk.SetBackend(clerk.NewBackend(&clerk.BackendConfig{
  HTTPClient: myCustomClient,
}))

domain.Create(ctx, &clerk.DomainParams{})
```

I'm facing the same dilemma as you. Is the global singleton-like backend worth the trouble? Will it actually prove flexible or not? I toyed around with different approaches, but DX-wise, I believe nothing beats the API described above.

How would the alternative look like?
```go
domain.NewClient(&clerk.ClientConfig).Create(ctx, &clerk.DomainParams{})
```

Or I imagine in your main.go file
```go
package main

type Server struct {
  domainClient *domain.Client
  organizationClient *organization.Client
}

func main() {
  // Instantiate your clients and your server struct to ListenAndServe
  clerkConfig := &clerk.BackendConfig{
    HTTPClient: myCustomClient,
  }
  srv := &Server{
    domainClient: domain.NewClient(clerkConfig),
    organizationClient: organization.NewClient(clerkConfig),
  }
}

// and then in your handler
func (srv *Server) CreateDomains(w, r) {
  dmn, err := srv.domainClient.Create(ctx, &clerk.DomainParams{})
}
```

Comparing this solution, to the `dmn, err := domain.Create()` example above, there's a clear DX benefit in my eyes.

What do you think?

clerk.go
```diff
@@ -0,0 +1,360 @@
...
+	"net/http"
+	"net/url"
+	"strconv"
+	"sync"
+	"time"
+)
+
+const (
+	sdkVersion   string = "v2.0.0"
+	clerkVersion string = "v1"
```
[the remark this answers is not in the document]
gkats: You mean `clerkAPIVersion` right? Updated.

clerk_test.go
georgepsarakis: It's a perfectly valid testing approach, although I think it's more usually followed in cases where handlers need to be tested. I just noted this in case we think we can avoid some boilerplate, that's all.

clerk.go
georgepsarakis: > Ability to stub out the client for tests

The use of a global variable here seems to be breaking test isolation, if they are running in parallel. So if I understand correctly we may not allowing dependency injection.

I think you can avoid the use of globals with defining an explicit setter for a default backend and provide constructors for dependency injection. I imagine the API being something like:


```go
package clerk

type Backend struct {
  httpClient *http.Client
}

var defaultBackend = &Backend{}
func SetDefaultBackend(cfg *BackendConfig) {
  // Resembles `SetBackend` now but mutates an existing struct
  defaultBackend.httpClient = cfg.HTTPClient
}

func DefaultBackend() *Backend {
  return defaultBackend
}

type DomainsClient struct {
  backend *Backend
}

func NewDomainsClient(backend *Backend) {}

var Domains = struct {
  backend *Backend
}{backend: defaultBackend}

func (d *Domains) Create(ctx context.Context, params *CreateDomainParams) (Domain, error) {}
```

```go
package main

func main() {
  // or this can happen on the user's package-level init()
  clerk.SetDefaultBackend(...)
}

func CreateDomains(ctx context.Context, name string) (..., error) {
  domain, err := clerk.Domains.Create(ctx, &clerk.DomainParams{Name: name})
}
```

Just an idea, I'm not sure what the exact end-user API needs to be.

clerk.go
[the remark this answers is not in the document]
georgepsarakis: Correct!

.github/workflows/ci.yml
```diff
@@ -0,0 +1,45 @@
...
+        with:
+          go-version: "1.21"
+      - name: Format
+        run: diff -u <(echo -n) <(gofmt -d -s .)
+      - name: Vet
+        run: go vet ./...
+      - name: Run linter
+        uses: golangci/golangci-lint-action@v3
+  test:
+    name: "Test: go v${{ matrix.go-version }}"
```
georgepsarakis, line 27: ❓ Just a note that you may want to retain the CI required check name. If I understand correctly all CI checks are optional now.

georgepsarakis (approved):
👏 Approving since this is foundational work in order to move things forward! There seems to be a failure for Go 1.17 that might be worth looking into.

clerk.go
```diff
@@ -0,0 +1,363 @@
...
+	"net/http"
+	"net/url"
+	"strconv"
+	"sync"
+	"time"
+)
+
+const (
+	sdkVersion      string = "v2.0.0"
+	clerkAPIVersion string = "v1"
```
alex-ntousias, line 22: I believe this is not needed anymore, given our plan for API versioning.

clerk.go
```diff
@@ -0,0 +1,363 @@
...
+)
+
+const (
+	// APIURL is the base URL for the Clerk API.
+	APIURL string = "https://api.clerk.com"
```
alex-ntousias, line 27: I know it's the Go convention, but `APIURL` looks sooo strange!

alex-ntousias (approved):
I really ❤️ where this is going! Great work @gkats .
@georgepsarakis raised some very valid concerns/questions. Other than that, looks great to me! :shipit:

clerk.go
gkats: Resolved the discussion offline. We'll go ahead with the current design until there's more strong evidence towards change.

clerk.go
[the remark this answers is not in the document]
gkats: Made the variable non-exported and we now offer a `clerk.SetKey("sk_live_123")` method to configure your API key.

clerk.go
gkats: Renamed. :smile:

clerk.go
gkats: :+1: When we're done with API versioning I'll revisit. Leaving this mostly as a placeholder for now.

PR #208, opened 2024-01-31
feat: Domains API operations

gkats:
## Type of change

## Description

Added types for Domains. 

Added a new package for Domains API operations. Supports List, Create, Update and Delete.

### DIFF
--- a/clerk.go
+++ b/clerk.go
@@ -258,7 +258,7 @@ func (b *defaultBackend) do(req *http.Request, params Queryable, setter Response
 func setRequestBody(req *http.Request, params Queryable) error {
 	// GET requests don't have a body, but we will pass the params
 	// in the query string.
-	if req.Method == http.MethodGet {
+	if req.Method == http.MethodGet && params != nil {
 		q := req.URL.Query()
 		params.Add(q)
 		req.URL.RawQuery = q.Encode()
--- a/domain.go
+++ b/domain.go
@@ -1,0 +1,25 @@
+package clerk
+
+type Domain struct {
+	APIResource
+	ID                string        `json:"id"`
+	Object            string        `json:"object"`
+	Name              string        `json:"name"`
+	IsSatellite       bool          `json:"is_satellite"`
+	FrontendAPIURL    string        `json:"frontend_api_url"`
+	AccountPortalURL  *string       `json:"accounts_portal_url,omitempty"`
+	ProxyURL          *string       `json:"proxy_url,omitempty"`
+	CNAMETargets      []CNAMETarget `json:"cname_targets,omitempty"`
+	DevelopmentOrigin string        `json:"development_origin"`
+}
+
+type CNAMETarget struct {
+	Host  string `json:"host"`
+	Value string `json:"value"`
+}
+
+type DomainList struct {
+	APIResource
+	Domains    []*Domain `json:"data"`
+	TotalCount int64     `json:"total_count"`
+}
--- a/domain/domain.go
+++ b/domain/domain.go
@@ -1,0 +1,73 @@
+// Package domain provides the Domains API.
+package domain
+
+import (
+	"context"
+	"net/http"
+	"net/url"
+
+	"github.com/clerk/clerk-sdk-go/v2"
+)
+
+const path = "/domains"
+
+type CreateParams struct {
+	clerk.APIParams
+	Name        *string `json:"name,omitempty"`
+	ProxyURL    *string `json:"proxy_url,omitempty"`
+	IsSatellite *bool   `json:"is_satellite,omitempty"`
+}
+
+// Create creates a new domain.
+func Create(ctx context.Context, params *CreateParams) (*clerk.Domain, error) {
+	req := clerk.NewAPIRequest(http.MethodPost, path)
+	req.SetParams(params)
+
+	domain := &clerk.Domain{}
+	err := clerk.GetBackend().Call(ctx, req, domain)
+	return domain, err
+}
+
+type UpdateParams struct {
+	clerk.APIParams
+	Name     *string `json:"name,omitempty"`
+	ProxyURL *string `json:"proxy_url,omitempty"`
+}
+
+// Update updates a domain's properties.
+func Update(ctx context.Context, id string, params *UpdateParams) (*clerk.Domain, error) {
+	path, err := url.JoinPath(path, id)
+	if err != nil {
+		return nil, err
+	}
+	req := clerk.NewAPIRequest(http.MethodPatch, path)
+	req.SetParams(params)
+
+	domain := &clerk.Domain{}
+	err = clerk.GetBackend().Call(ctx, req, domain)
+	return domain, err
+}
+
+// Delete removes a domain.
+func Delete(ctx context.Context, id string) (*clerk.DeletedResource, error) {
+	path, err := url.JoinPath(path, id)
+	if err != nil {
+		return nil, err
+	}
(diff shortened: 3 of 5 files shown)


domain.go
```diff
@@ -0,0 +1,25 @@
+package clerk
```
georgepsarakis, line 1: ❓ Couldn't we use the `domain` package here instead?

domain/domain.go
```diff
@@ -0,0 +1,73 @@
...
+	err = clerk.GetBackend().Call(ctx, req, domain)
+	return domain, err
+}
+
+type ListParams struct {
+	clerk.APIParams
+}
+
+// List returns a list of domains
+func List(ctx context.Context, params *ListParams) (*clerk.DomainList, error) {
```
georgepsarakis, line 68: 🔧 If there's not a strong reason to use pointers, I think we should avoid them where possible, as they make garbage collection more difficult and the risk of accidental mutations increases.

domain/domain_test.go
```diff
@@ -0,0 +1,193 @@
...
+	// If set, we'll assert that the request body
+	// matches.
+	in json.RawMessage
+	// If set, we'll assert the request path matches.
+	path string
+	// If set, we'll assert that the request method matches.
+	method string
+}
+
+func (rt *mockRoundTripper) RoundTrip(r *http.Request) (*http.Response, error) {
```
georgepsarakis, line 169: ❓ This stub should be used everywhere in the project, right? Perhaps consider moving to centralized tetsing utilities?

domain/domain_test.go
```diff
@@ -0,0 +1,193 @@
...
+	if rt.status == 0 {
```
georgepsarakis, line 170: 💡 Mostly a suggestion, it might be easier to use a constructor for initializations: `newMockRoundTripper`

domain/domain_test.go
```diff
@@ -0,0 +1,193 @@
...
+		rt.status = http.StatusOK
+	}
+
+	if rt.method != "" {
+		require.Equal(rt.T, rt.method, r.Method)
+	}
+	if rt.path != "" {
+		require.Equal(rt.T, rt.path, r.URL.Path)
+	}
+	if rt.in != nil {
```
georgepsarakis, line 180: ❓ Shouldn't we assert that the body is empty if `rt.in == nil`?

domain/domain_test.go
```diff
@@ -0,0 +1,193 @@
...
+	// If set, we'll assert that the request method matches.
+	method string
+}
+
+func (rt *mockRoundTripper) RoundTrip(r *http.Request) (*http.Response, error) {
+	if rt.status == 0 {
+		rt.status = http.StatusOK
+	}
+
+	if rt.method != "" {
```
georgepsarakis, line 174: 🔧 I don't think you need this check?

domain/domain_test.go
```diff
@@ -0,0 +1,193 @@
...
+		require.Equal(rt.T, rt.method, r.Method)
+	}
+	if rt.path != "" {
```
georgepsarakis, line 177: 🔧 Probably redundant check here too, I imagine that all requests have a path that we want to assert.

georgepsarakis (commented):
👏

domain.go
gkats: We could, but I'm afraid that at some point we might need this type from another package. Imagine an Instance which has a nested Domain.

I believe it's safer to have these types for "models" in the `clerk` package so we don't run into any circular dependecy problem in the future.

domain/domain_test.go
gkats: Yup, I already have. See https://github.com/clerk/clerk-sdk-go/pull/209/files#diff-edaf0121fb36b9609248644a9f18247c562847dbb19fa4bcbf5c99e03eb7b921R1

domain/domain_test.go
gkats: Not sure about that, I left it on purpose like this so we don't always have to define a body.

I mean, there are some tests where you care about the request body (see the Create test) and others where you probable don't (see the Create_Error test).

I'm happy to change this as more use-cases come along, but for now it does the job.

domain/domain_test.go
gkats: I do, because I don't always want to have to define the method. For some tests, I simply don't care.

domain/domain_test.go
gkats: Yes, but we might not want to assert on the path in all tests.

domain.go
georgepsarakis: I see, I thought that we might want to reuse structs. Do you think that if we introduced a `models/<service>` kind of structure, we might run into circular dependency problems? Just thinking out loud here:

```go
import (
  domainmodels "github.com/clerk/clerk-sdk-go/v2/models/domains"
)
```

clerk.go
```diff
@@ -258,7 +258,7 @@ func (b *defaultBackend) do(req *http.Request, params Queryable, setter Response
 func setRequestBody(req *http.Request, params Queryable) error {
 	// GET requests don't have a body, but we will pass the params
 	// in the query string.
-	if req.Method == http.MethodGet {
+	if req.Method == http.MethodGet && params != nil {
 		q := req.URL.Query()
 		params.Add(q)
```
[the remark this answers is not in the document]
gkats: Great point! I'll update the interface and the logic. I'd rather do it on a separate PR. Do you think that would be a problem?

georgepsarakis (approved):

clerk.go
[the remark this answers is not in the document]
georgepsarakis: Not a problem at all, by all means proceed with merging 🚀

domain.go
gkats: Thanks for the suggestion! Your solution would certainly avoid circular dependency, but isn't it a bit too much packaging?

Is there any reason why using the clerk package might be problematic?

domain/domain_test.go
```diff
@@ -0,0 +1,193 @@
...
+		require.Equal(rt.T, rt.path, r.URL.Path)
+	}
+	if rt.in != nil {
+		body, err := io.ReadAll(r.Body)
+		if err != nil {
+			return nil, err
+		}
+		defer r.Body.Close()
+		require.JSONEq(rt.T, string(rt.in), string(body))
```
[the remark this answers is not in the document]
gkats: If I want to use `JSONEq` I have to cast. 

https://pkg.go.dev/github.com/stretchr/testify@v1.8.4/assert#JSONEq

There are alternative ways to compare, but I guess this is fine as is.

domain.go
georgepsarakis: It's not problematic, it's only that packages provide a native namespace and the import aliases provide some more naming flexibility for someone that uses them. Names also become more uniform.

```go
package domain

// Debatable name :)
type Object struct{}
type List struct {
  Data []Object `json:"data"`
}
```

```go
import (
  domainmodels "github.com/clerk/clerk-sdk-go/v2/models/domains"
) 

func HandleDomains() {
  var domainList domainmodels.List
  //...
}
```

Just an idea here, not insisting on this, it might be overly verbose/complicated for some reason. Noting now, since it would be a breaking change afterwards!

domain/domain_test.go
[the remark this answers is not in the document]
georgepsarakis: Pfff sorry about that, I should've checked the signature first, I was for some reason certain that it uses `[]byte` 😞

PR #209, opened 2024-01-31
feat: Actor token API operations

gkats:
## Type of change

## Description

Added the types and operations for actor tokens. Create and Revoke are supported.

### DIFF
--- a/actortoken/actor_token.go
+++ b/actortoken/actor_token.go
@@ -1,0 +1,42 @@
+// Package actortoken provides the Actor Tokens API.
+package actortoken
+
+import (
+	"context"
+	"encoding/json"
+	"net/http"
+	"net/url"
+
+	"github.com/clerk/clerk-sdk-go/v2"
+)
+
+const path = "/actor_tokens"
+
+type CreateParams struct {
+	clerk.APIParams
+	UserID                      *string         `json:"user_id,omitempty"`
+	Actor                       json.RawMessage `json:"actor,omitempty"`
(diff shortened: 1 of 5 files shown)


actortoken/actor_token.go
```diff
@@ -0,0 +1,42 @@
...
+// Revoke revokes a pending actor token.
+func Revoke(ctx context.Context, id string) (*clerk.ActorToken, error) {
+	path, err := url.JoinPath(path, id, "revoke")
+	if err != nil {
+		return nil, err
+	}
+	req := clerk.NewAPIRequest(http.MethodPost, path)
+	token := &clerk.ActorToken{}
+	err = clerk.GetBackend().Call(ctx, req, token)
+	return token, err
```
georgepsarakis, line 41: ❓ Do you want to return `nil` or error or a pointer to empty struct? Because as it is now, it seems more suitable to return the empty value instead of a pointer.

georgepsarakis (approved):

actortoken/actor_token.go
gkats: You mean because we return either nil or the pointer and we should be consistent and always return one or the other in case of error?

georgepsarakis: Usually with pointers as return values we see in the code something like:

```go
if err != nil {
  return nil, err
}
```

Here we return a pointer to empty struct which is not `nil`. Technically, non-nil error values should always take precedence at the call site. I am personally in favor for returning values not pointers:

```go
return clerk.ActorToken{}, err
```

PR #210, opened 2024-02-01
fix: Params interface

gkats:
## Type of change

## Description

Changed the Queryable interface to Params. Instead of operating on a url.Values object, implementing types should return a url.Values object. When we set the params to the querystring, we'll use the new interface method.

Added a custom implementation to join URL paths. The new implementation will preserve any query parameters in the URL. We can now replace url.JoinPath, which required us to use Go v1.19 and upwards.

### DIFF
--- a/clerk.go
+++ b/clerk.go
@@ -12,7 +12,9 @@ import (
 	"io"
 	"net/http"
 	"net/url"
+	"regexp"
 	"strconv"
+	"strings"
 	"sync"
 	"time"
 )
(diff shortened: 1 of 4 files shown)


clerk.go
```diff
@@ -277,7 +279,24 @@ func setRequestBody(req *http.Request, params Queryable) error {
...
+// always win.
+func setRequestQuery(req *http.Request, params Params) {
+	if params == nil {
+		return
+	}
+	q := req.URL.Query()
+	paramsQuery := params.ToQuery()
+	for k, values := range paramsQuery {
+		for _, v := range values {
+			q.Set(k, v)
```
georgepsarakis, line 293: ❓ `Set` or `Add`? As long as we don't consider a possible future case where we must set a query parameter at the SDK level (I don't think there's one), should be 100% fine though.

georgepsarakis (approved):

Issue #202, opened 2024-01-15
WithCustomClaims in Verify token options should accept a type instead of pointer


vijathanga:
Currently `WithCustomClaims` function in `token_option.go` accepts a pointer to a struct variable. So we have to define and pass the struct pointer when initializing the `WithSessionV2` middleware. The same struct pointer is used for unmarshalling the custom claims for all incoming requests. This can cause consistency issues when multiple requests are processed at the same time.

This can be solved by either
 1. Accept a Custom claim struct type in `WithCustomClaims` function and dynamically create, populate the struct and set the value in context
 2. Provide a function hook for dealing with custom claims

georgepsarakis:
Hello @vijathanga ! Sorry that you encountered this issue and thanks for reporting. Your observation is correct:

> This can cause consistency issues when multiple requests are processed at the same time.

The `WithCustomClaims` option was designed for direct calls to `VerifyToken` since the custom claims are assigned directly to the referenced struct instance which would be constructed before each `VerifyToken` call:

https://github.com/clerk/clerk-sdk-go/blob/2f041fef8d673473d84d5b8360b0650d43b11ae3/clerk/clerk.go#L51

> Accept a Custom claim struct type in WithCustomClaims function

Do you perhaps mean a constructor function for new objects that will be populated with the custom claims? Something like:

```go
WithSessionV2(client, WithSessionClaimsFunc(func() interface{} {
  return &MyCustomClaims{}
})
```

> populate the struct and set the value in context

Would exposing the custom claims as a field in the `SessionClaims` struct that's already available in the context work instead?

https://github.com/clerk/clerk-sdk-go/blob/2f041fef8d673473d84d5b8360b0650d43b11ae3/clerk/middleware_v2.go#L38

So to sum up the approach assuming a custom struct `MyCustomClaims`:

```go
client, err := clerk.NewClient(apiKey)
// Configure the middleware
clerk.WithSessionV2(client, WithSessionClaimsFunc(func() interface{} {
  return &MyCustomClaims{}
})

// In the request
sessionClaims, ok := clerk.SessionFromContext(req.Context())
if !ok {
  // unauthenticated
}
customClaims, ok := sessionClaims.Custom.(*MyCustomClaims)
if ok {
  // use customClaims
}
```

Mostly asking if the final snippet would address your use case, this change will require some additional design considerations from our side!

vijathanga:
Hi @georgepsarakis, I appreciate the quick response.

[...]

PR #214, opened 2024-02-05
feat: HTTP middleware

gkats:
HTTP middleware for enhancing http.Handler with Clerk authentication. The http package provides two middleware:

  - WithHeaderAuthentication checks the Authorization header for a valid JWT and sets the active session claims in the request context.
  - RequireHeaderAuthentication responds with 403 Forbidden and halts the handler execution chain if it's unable to detect valid session claims. 

Example usage:
```go
ctx := context.Background()
rootHandler := func(w http.ResponseWriter, r *http.Request) {
  claims, _ := clerk.SessionClaimsFromContext(r.Context())
  w.Write([]byte(fmt.Sprintf(`{"user_id":"%s"}`, claims.UserID)))
}
// Adds active session claims to the request context
http.HandleFunc("/", clerkhttp.WithHeaderAuthentication(ctx)(rootHandler)
```

Please note that cookie based authentication is not supported. The SDK v1 middleware supported [cookie authentication](https://github.com/clerk/clerk-sdk-go/blob/afee4cadefab3bc98a1eb87c90cc0a41a3dc586d/clerk/middleware_v2.go#L96) as well.

Added a new jwt package which implements the core function to validate a JWT. The function name is Verify and is used by the middleware to validate the Bearer JWT. The middleware above uses the Verify() method internally. 

Added a new jwk package which provides operations for the JSON Web Keys API. We need to be able to fetch the instance JWK in order to validate and parse the Bearer token.

Added a dependency to go-jose/v3 package for performing operations on JWTs.


http/middleware.go
```diff
@@ -0,0 +1,150 @@
...
+		params.IsSatellite = isSatellite
+		return nil
+	}
+}
+
+// JSONWebKey allows to provide a custom JSON Web Key based on which
+// the authorization JWT will be verified.
+func JSONWebKey(key string) AuthorizationOption {
+	return func(params *AuthorizationParams) error {
+		// From the Clerk docs: "Note that the JWT Verification key is not in
```
georgepsarakis, line 137: 🔧 Might be worth adding a unit test here, even for illustrating the use a bit better.

http/middleware_test.go
```diff
@@ -0,0 +1,67 @@
...
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/clerk/clerk-sdk-go/v2"
+	"github.com/stretchr/testify/require"
+)
+
+func TestWithHeaderAuthentication_InvalidAuthorization(t *testing.T) {
```
georgepsarakis, line 13: ❓ Don't you need a successful test case as well?

jwt.go
```diff
@@ -0,0 +1,67 @@
...
+import (
+	"context"
+	"encoding/json"
+
+	"github.com/go-jose/go-jose/v3/jwt"
+)
+
+type key string
+
+const activeSessionClaims = key("activeSessionClaims")
```
georgepsarakis, line 12: 💡 We can prefix/namespace the context key with `clerk` so the scope is even clearer if someone inspects its contents.

jwt.go
```diff
@@ -0,0 +1,67 @@
...
+	ActiveOrganizationRole        string          `json:"org_role"`
+	ActiveOrganizationPermissions []string        `json:"org_permissions"`
+	Actor                         json.RawMessage `json:"act,omitempty"`
+}
+
+// HasPermission checks if the session claims contain the provided
+// organization permission.
+// Use this helper to check if a user has the specific permission in
+// the active organization.
+func (s *SessionClaims) HasPermission(permission string) bool {
```
georgepsarakis, line 43: ❓ Aren't there some unit tests for these methods? Shouldn't we transfer them as well? Perhaps I've missed it though.

jwt/jwt.go
```diff
@@ -0,0 +1,156 @@
...
+	azp := make(map[string]struct{})
+	for _, p := range parties {
+		azp[p] = struct{}{}
+	}
+	params.authorizedParties = azp
+}
+
+// Verify verifies a Clerk session JWT and returns the parsed
+// clerk.SessionClaims.
+func Verify(ctx context.Context, params *VerifyParams) (*clerk.SessionClaims, error) {
```
georgepsarakis, line 47: ❓ Is this copied verbatim from the previous package version?

georgepsarakis (commented):
👏 Just a few questions, overall looks fine to me.

jwk.go
```diff
@@ -0,0 +1,57 @@
...
+type JSONWebKey struct {
+	APIResource
+	jose.JSONWebKey
+	Key       any    `json:"key"`
+	KeyID     string `json:"kid"`
+	Algorithm string `json:"alg"`
+	Use       string `json:"use"`
+}
+
+func (k *JSONWebKey) UnmarshalJSON(data []byte) error {
```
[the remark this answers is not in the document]
gkats: I wanted to avoid using a go-jose type and instead make sure we use only our own package's (clerk) types everywhere.

There's no reason to expose types that are meant for internal use anywhere. We unfortunately do this in v1 of our SDK.

The go-jose library is able to decode JSON web keys and we can leverage that to get a response from our Backend API GET /v1/jwks endpoint and translate it to our own `clerk.JSONWebKeySet` and effectively `clerk.JSONWebKey`.

Unfortunately, it looks like extracting the key ID and key from a JWK response is not that straightforward. Depending on the algorithm that's used the JWK response might have different fields set, each with different meaning.

The go-jose library [handles unmarshaling nicely](https://github.com/go-jose/go-jose/blob/47edce0854d533ac27795c9befd90b1f7ef87554/jwk.go#L166), but its [raw unmarshaled type is unexported](https://github.com/go-jose/go-jose/blob/47edce0854d533ac27795c9befd90b1f7ef87554/jwk.go#L42). 

So, what I thought we could do is use the jose.JSONWebKey type to unmarshal the raw JSON and then copy over the resulting fields to our type.

I initially tried type aliasing but it didn't work as expected.

Hope the above makes the decision clearer. Let me know if I should offer more details! Perhaps I can add a comment explaining what's going on here?

jwk/jwk.go
```diff
@@ -0,0 +1,25 @@
+// Package jwk provides the JSON Web Tokens API.
+package jwk
```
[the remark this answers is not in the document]
gkats: Indeed. I've opted for a singular term for all packages so far and thought I'd continue the same pattern. Happy to change it though if you think it's a problem.

http/middleware.go
gkats: My intention was to add an integration test for the options, but setting the tests up took more time than expected. I'm planning to get back to this once I find some time.

http/middleware_test.go
gkats: Yes, setting up tests for token verification is cumbersome, but I'll get back to this. I'll open a ticket to track this.

jwt.go
gkats: :+1: Added new tests.

jwt.go
gkats: Great idea! Updated.

jwk.go
```diff
@@ -0,0 +1,57 @@
...
+		return nil, fmt.Errorf("invalid PEM-encoded block")
+	}
+
+	if block.Type != "PUBLIC KEY" {
+		return nil, fmt.Errorf("invalid key type, expected a public key")
+	}
+
+	rsaPublicKey, err := x509.ParsePKIXPublicKey(block.Bytes)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse public key: %v", err)
```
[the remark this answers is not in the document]
gkats: :+1: You are right, copy-pasted it from v1. Changed it to wrap the error with %w instead.

http/middleware_test.go
kostaspt, line 13: 🙃  Looks like there's a mix-up between Authentication and Authorization, even in the PR description. But Authorization seems to be the right term for this case.

```suggestion
func TestWithHeaderAuthorization_InvalidAuthorization(t *testing.T) {
```

gkats: Updated!

jwk/jwk.go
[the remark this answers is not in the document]
agis: Since "jwks" is not a plural term but an acronym, I think the singular/plural rationale (which I'm all for!) does not apply here, right? If so, I tend to agree with George's suggestion.

jwk/jwk.go
```diff
@@ -0,0 +1,25 @@
...
+}
+
+// Get retrieves a JSON Web Key set.
+func Get(ctx context.Context, params *GetParams) (*clerk.JSONWebKeySet, error) {
+	req := clerk.NewAPIRequest(http.MethodGet, path)
+	req.SetParams(params)
+
+	set := &clerk.JSONWebKeySet{}
+	err := clerk.GetBackend().Call(ctx, req, set)
+	return set, err
```
agis, line 24: ❓ Wouldn't it be more idiomatic to return `nil` if there's an error?

jwk.go
```diff
@@ -0,0 +1,57 @@
...
+import (
+	"crypto/x509"
+	"encoding/pem"
+	"fmt"
+
+	"github.com/go-jose/go-jose/v3"
+)
+
+type JSONWebKeySet struct {
+	APIResource
```
agis, line 12: ❓ What is this for?

jwk.go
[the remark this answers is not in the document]
agis: > I wanted to avoid using a go-jose type and instead make sure we use only our own package's (clerk) types everywhere. There's no reason to expose types that are meant for internal use anywhere. We unfortunately [do this](https://github.com/clerk/clerk-sdk-go/blob/8bd66b16f28be05ad4377d3f24d2ace2c65d5fb3/clerk/tokens.go#L43) in v1 of our SDK.

Very good call. This provides us with flexibility to change the underlying library (go-jose) without it being a breaking change.

jwk.go
```diff
@@ -0,0 +1,57 @@
...
+	Keys []JSONWebKey `json:"keys"`
+}
+
+type JSONWebKey struct {
+	APIResource
+	jose.JSONWebKey
```
agis, line 18: 🔧 Since we don't want to expose jose, shouldn't we avoid embedding this and instead wrap it with our own implementation?

jwk/jwk.go
```diff
@@ -0,0 +1,25 @@
+// Package jwk provides the JSON Web Tokens API.
```
agis, line 1: 🔧 How about something like:

> Package jwks provides access to the JWKs endpoint.
>
> See https://clerk.com/docs/reference/backend-api/tag/JWKS

http/middleware.go
```diff
@@ -0,0 +1,150 @@
...
+
+			params := &AuthorizationParams{}
+			for _, opt := range opts {
+				err = opt(params)
+				if err != nil {
+					w.WriteHeader(http.StatusUnauthorized)
+					return
+				}
+			}
+			params.Token = token
```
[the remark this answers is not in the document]
gkats: I wanted to make sure it's not accidentally overriden by the options, but I guess you're right.

jwt/jwt.go
```diff
@@ -0,0 +1,156 @@
...
+			return nil, fmt.Errorf("invalid authorized party %s", claims.AuthorizedParty)
+		}
+	}
+
+	return claims, nil
+}
+
+func getJWK(ctx context.Context, kid string) (*clerk.JSONWebKey, error) {
+	// TODO Avoid multiple requests by caching results for the same
+	// instance.
```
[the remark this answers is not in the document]
gkats: :+1: I already have an implementation ready, waiting for this PR to get merged before I push it. :smile: 

We're not releasing anything yet, there's still more ground we have to cover.

http/middleware.go
```diff
@@ -0,0 +1,150 @@
...
+	jwt "github.com/clerk/clerk-sdk-go/v2/jwt"
+)
+
+// RequireAuthorization attempts to get Clerk authorization claims
+// from the http.Request context and responds with status  403 Forbidden
+// if it fails to verify the claims.
+func RequireHeaderAuthorization(ctx context.Context, opts ...AuthorizationOption) func(http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
+		return WithHeaderAuthorization(ctx, opts...)(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			claims, ok := clerk.SessionClaimsFromContext(r.Context())
```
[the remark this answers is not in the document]
gkats: Good question, I had the same doubts.

 I left the `ctx` context for possible future usage. 

Although it's not actually needed, if we don't add it now and we need it in a future version, we'll break the API.

Do you think it's better to remove it?

http/middleware.go
```diff
@@ -0,0 +1,150 @@
...
+}
+
+// JSONWebKey allows to provide a custom JSON Web Key based on which
+// the authorization JWT will be verified.
+func JSONWebKey(key string) AuthorizationOption {
+	return func(params *AuthorizationParams) error {
+		// From the Clerk docs: "Note that the JWT Verification key is not in
+		// PEM format, the header and footer are missing, in order to be shorter
+		// and single-line for easier setup."
+		if !strings.HasPrefix(key, "-----BEGIN") {
```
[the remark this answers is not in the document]
gkats: Not sure about this in general. 

If it was my decision, I think we should always require the full key here. 

I simply carried over the logic from v1, but I could add a check for the suffix as well.

jwt/jwt.go
```diff
@@ -0,0 +1,156 @@
...
+	"github.com/go-jose/go-jose/v3/jwt"
+)
+
+type VerifyParams struct {
+	// Token is the JWT that will be verified.
+	Token string
+	// JWK is a custom JSON Web Key that can be provided to skip
+	// fetching one.
+	JWK          *clerk.JSONWebKey
+	CustomClaims any
```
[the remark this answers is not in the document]
gkats: I'm planning to revisit this in order to fix #202, that's why it's the only field that doesn't have a comment yet. :smile:

jwt/jwt.go
```diff
@@ -0,0 +1,156 @@
...
+	iss := claims.Issuer
+	if params.ProxyURL != nil && *params.ProxyURL != "" {
+		iss = *params.ProxyURL
+	}
+	// Non-satellite domains must validate the issuer.
+	if !params.IsSatellite && !isValidIssuer(iss) {
+		return nil, fmt.Errorf("invalid issuer %s", iss)
+	}
+
+	if claims.AuthorizedParty != "" && len(params.authorizedParties) > 0 {
```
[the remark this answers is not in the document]
gkats: I'm not sure, but open to any suggestion.

The check is copied over from v1.

See https://github.com/clerk/clerk-sdk-go/blob/8bd66b16f28be05ad4377d3f24d2ace2c65d5fb3/clerk/tokens.go#L102

http/middleware.go
```diff
@@ -0,0 +1,150 @@
...
+	"context"
+	"net/http"
+	"strings"
+	"time"
+
+	"github.com/clerk/clerk-sdk-go/v2"
+	jwt "github.com/clerk/clerk-sdk-go/v2/jwt"
+)
+
+// RequireAuthorization attempts to get Clerk authorization claims
```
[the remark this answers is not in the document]
gkats: :+1: Nice, updated!

http/middleware.go
```diff
@@ -0,0 +1,150 @@
...
+			}
+			params.Token = token
+			claims, err := jwt.Verify(ctx, &params.VerifyParams)
+			if err != nil {
+				w.WriteHeader(http.StatusUnauthorized)
+				return
+			}
+
+			// Token was verified. Add the session claims to the request context
+			sessionCtx := clerk.ContextWithSessionClaims(r.Context(), claims)
```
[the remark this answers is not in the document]
gkats: Updated to `newCtx`.

http/middleware.go
```diff
@@ -0,0 +1,150 @@
...
+// See https://clerk.com/docs/advanced-usage/satellite-domains
+func Satellite(isSatellite bool) AuthorizationOption {
+	return func(params *AuthorizationParams) error {
+		params.IsSatellite = isSatellite
+		return nil
+	}
+}
+
+// JSONWebKey allows to provide a custom JSON Web Key based on which
+// the authorization JWT will be verified.
```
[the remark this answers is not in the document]
gkats: Enhanced the comment with more info.

jwk.go
gkats: Well, yes! Great catch, I missed this. Used an unexported field instead.

jwk/jwk.go
gkats: :+1: Updated.

jwt/jwt.go
```diff
@@ -0,0 +1,156 @@
...
+		strings.Contains(iss, ".clerk.accounts")
+}
+
+type DecodeParams struct {
+	Token string
+}
+
+// Decode decodes a JWT without verifying it. This is a quick way
+// to validate the format of a JSON web token, but the resulting Claims
+// should be considered untrusted.
```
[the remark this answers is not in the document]
gkats: Done.

jwk/jwk.go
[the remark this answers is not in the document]
gkats: Yes, initially I thought I'd treat this as a "JSON Web Key" API (hence jwk). 

I guess everybody else expects the "JSON Web Key Set" API, so I've changed it.

http/middleware.go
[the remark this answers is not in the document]
georgepsarakis: I'm not sure I can think of a good usage pattern here too, since the middleware setup doesn't happen within an operation context. Perhaps we should remove it?

http/middleware.go
```diff
@@ -0,0 +1,155 @@
...
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			authorization := strings.TrimSpace(r.Header.Get("Authorization"))
+			if authorization == "" {
+				next.ServeHTTP(w, r)
+				return
+			}
+
+			token := strings.TrimPrefix(authorization, "Bearer ")
+			_, err := jwt.Decode(ctx, &jwt.DecodeParams{Token: token})
```
georgepsarakis, line 45: 🔧 I think you need the Request context here:
```suggestion
			_, err := jwt.Decode(r.Context(), &jwt.DecodeParams{Token: token})
```

http/middleware.go
```diff
@@ -0,0 +1,155 @@
...
+func AuthorizedParty(parties ...string) AuthorizationOption {
+	return func(params *AuthorizationParams) error {
+		params.SetAuthorizedParties(parties...)
+		return nil
+	}
+}
+
+// CustomClaims allows to pass a type (e.g. struct), which will be populated with the token claims based on json tags.
+// You must pass a pointer for this option to work.
+func CustomClaims(claims any) AuthorizationOption {
```
georgepsarakis, line 92: ℹ️ I think you're already aware but just adding a reference for #202 before we solidify the option design!

http/middleware.go
```diff
@@ -0,0 +1,155 @@
...
+			params := &AuthorizationParams{}
+			for _, opt := range opts {
+				err = opt(params)
+				if err != nil {
+					w.WriteHeader(http.StatusUnauthorized)
+					return
+				}
+			}
+			params.Token = token
+			claims, err := jwt.Verify(ctx, &params.VerifyParams)
```
georgepsarakis, line 60: 🔧 Suggestion for a later PR to perhaps consider allowing client injection dependency here for testing.

jwk.go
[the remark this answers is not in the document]
georgepsarakis: I see, yes good call indeed. Now that the field is unexported the purpose is also much clearer!

jwt.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/go-jose/go-jose/v3/jwt"
+)
+
+type key string
+
+const clerkActiveSessionClaims = key("clerkActiveSessionClaims")
+
+// ContextWithSessionClaims returns a new context which includes the
+// active session claims.
+func ContextWithSessionClaims(ctx context.Context, value any) context.Context {
```
georgepsarakis, line 16: 🔧 Shouldn't the `value` type here be `*SessionClaims` 🤔 ?

jwt/jwt.go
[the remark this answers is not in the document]
georgepsarakis: Oh just saw your comment here, please ignore my comment above referencing the same issue 😬

jwt/jwt.go
```diff
@@ -0,0 +1,156 @@
...
+	allClaims := []any{claims}
+	if params.CustomClaims != nil {
+		allClaims = append(allClaims, params.CustomClaims)
+	}
+	err = parsedToken.Claims(jwk.Key, allClaims...)
+	if err != nil {
+		return nil, err
+	}
+
+	err = claims.Claims.ValidateWithLeeway(jwt.Expected{Time: time.Now()}, params.Leeway)
```
georgepsarakis, line 83: 🔧 
```suggestion
	err = claims.Claims.ValidateWithLeeway(jwt.Expected{Time: time.Now().UTC()}, params.Leeway)
```

georgepsarakis (approved):

jwt.go
gkats: :thinking: I'm not sure about that. TBD after we settle on custom type for the claims.

http/middleware.go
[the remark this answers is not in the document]
gkats: Removed the context from the arguments.

http/middleware.go
```diff
@@ -0,0 +1,154 @@
...
+	return func(params *AuthorizationParams) error {
+		params.Leeway = leeway
+		return nil
+	}
+}
+
+// ProxyURL can be used to set the URL that proxies the Clerk Frontend
+// API. Useful for proxy based setups.
+// See https://clerk.com/docs/advanced-usage/using-proxies
+func ProxyURL(proxyURL string) AuthorizationOption {
```
agis, line 114: 💡 Shall we validate that `proxyURL` is indeed a valid URL?

http/middleware.go
```diff
@@ -0,0 +1,154 @@
...
+	return func(params *AuthorizationParams) error {
+		params.ProxyURL = clerk.String(proxyURL)
+		return nil
+	}
+}
+
+// Satellite can be used to signify that the authorization happens
+// on a satellite domain.
+// See https://clerk.com/docs/advanced-usage/satellite-domains
+func Satellite(isSatellite bool) AuthorizationOption {
```
agis, line 124: ❓ Is there any reason for someone to call this as `Satellite(false)`? If not, then perhaps this should be simplified to `Satellite()` (and always set `params.IsSatellite = true`).

http/middleware.go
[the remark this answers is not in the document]
agis: It seems to me that it would be more fool-proof than what v1 does, but then again I understand if you decide it's not even worth it, since it's an edge case. Up to you! :)

jwt/jwt.go
```diff
@@ -0,0 +1,156 @@
...
+	return nil, fmt.Errorf("no jwk key found for kid %s", kid)
+}
+
+func isValidIssuer(iss string) bool {
+	return strings.HasPrefix(iss, "https://clerk.") ||
+		strings.Contains(iss, ".clerk.accounts")
+}
+
+type DecodeParams struct {
+	Token string
```
agis, line 127: ❓ What's the reason for wrapping `Token` in a struct? Is it for providing us a way to add more options without breaking compatibility (i.e. keeping the function signature intact)?

If so, I'm having a hard time imagining something that `Decode` would need to be extended with, that a functional option wouldn't be a fit for. In other words, perhaps it's perfectly fine to go with `Decode(ctx, string)` and in the future extend with functional options if needed? The reason I'm suggesting this is because the latter seems a simpler/easier to use API to me.

jwt/jwt.go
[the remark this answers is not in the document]
agis: Interesting. That's maybe worthy of more investigation, but it's out of scope of this PR. I'll create an internal ticket to investigate further.

agis (approved):
Approving for expediency, I don't see any blockers here since we're not releasing yet. Awesome work! 💯

jwt/jwt.go
gkats: Yes, we're trying to use custom types for method arguments so that we can easily extend them in the future without introducing breaking changes in the API.

Issue #216, opened 2024-02-05
Issue with UpdateUserMetadata


krishnadubagunta:
I found this issue similar to what I was facing. https://github.com/clerk/clerk-sdk-go/issues/109 

I couldn't update the metadata, it fails on awaiting headers for some reason.  I made sure my keys are current, and are passing into the request. Here what I get, 

```
Patch "https://api.clerk.dev/v1/users/<userId>/metadata": context deadline exceeded (Client.Timeout exceeded while awaiting headers)
```

For this code

```go
updatedMetadata := map[string]string{
	"artist_id": strconv.FormatInt(ar.ID, 10),
	"role":      ar.Role,
}
mar, _ = json.Marshal(updatedMetadata)
retry := 5
for i := 0; i < retry; i++ {
	// err := updateUserMetadata(webhookArtist.Data.ID, mar)
	user, err := client.Users().UpdateMetadata(webhookArtist.Data.ID, &clerk.UpdateUserMetadata{
		PublicMetadata: mar,
	})
	if err != nil {
		log.Println(err)
		i += 1
	} else {
		log.Println("Worked ", user)
		break
	}
}
```
and when I give the same data into the http request, it updates metadata, like so
```go
url := fmt.Sprintf("https://api.clerk.dev/v1/users/%s/metadata", userID)
method := "PATCH"

payload := map[string]interface{}{
	"public_metadata": message,
}

j, err := json.Marshal(payload)
if err != nil {
	return err
}

client := &http.Client{}

req, err := http.NewRequest(method, url, bytes.NewReader(j))
if err != nil {
	return err
}

req.Header.Add("Authorization", fmt.Sprintf("Bearer %s", "token"))
req.Header.Add("Content-Type", "application/json")

res, err := client.Do(req)
if err != nil {
	return err
}

return res.Body.Close()
```
Is there something that I'm missing?

gkats:
Hello @krishnadubagunta thanks for reporting the issue.

Is the problem happening constantly? The error message suggests that it might be a temporary issue.

krishnadubagunta:
@gkats I've only began to use clerk 3 days ago, and this is constantly happening.

gkats:
> I've only began to use clerk 3 days ago, and this is constantly happening.

Thanks for the reply. Seems to be a bigger problem then. We'll look into it.

krishnadubagunta:
> > I've only began to use clerk 3 days ago, and this is constantly happening.
> 
[...]

PR #220, opened 2024-02-06
Increase timeout

krishnadubagunta:
Regarding this issue https://github.com/clerk/clerk-sdk-go/issues/216 that I opened.
After much trouble, out of desperation, I increased timeout by 1 more second and it worked?. 

I'm not sure what caused this delay, I would guess 5 seconds is too much for key verification, and I had to increase it by 1. do you think this was my work env's fault?

I'm able to fetch the user data and update records by this update.

### DIFF
--- a/clerk/clerk.go
+++ b/clerk/clerk.go
@@ -41,7 +41,7 @@ const (
 	JWTTemplatesUrl    = "jwt_templates"
 )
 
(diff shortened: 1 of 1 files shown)


krishnadubagunta:
Thanks for the help @gkats 
Closing this issue in favor of the resolving comment on issue https://github.com/clerk/clerk-sdk-go/issues/216

PR #221, opened 2024-02-06
feat: Custom User-Agent strings

gkats:
Added a new configuration option for our Backend so that users can specify a suffix for the User-Agent string that's included in every request headers.

### DIFF
--- a/clerk.go
+++ b/clerk.go
@@ -146,6 +146,9 @@ type BackendConfig struct {
 	// URL is the base URL to use for API endpoints.
 	// If it's not set, the default value for the Backend will be used.
 	URL *string
+	// UserAgentSuffix is appended to the User-Agent string on every
+	// request. Useful for support or debugging purposes.
+	UserAgentSuffix *string
 }
 
 // NewBackend returns a default backend implementation with the
@@ -160,8 +163,9 @@ func NewBackend(config *BackendConfig) Backend {
 		config.URL = String(APIURL)
 	}
 	return &defaultBackend{
-		HTTPClient: config.HTTPClient,
-		URL:        *config.URL,
+		HTTPClient:      config.HTTPClient,
+		URL:             *config.URL,
+		UserAgentSuffix: config.UserAgentSuffix,
 	}
 }
 
@@ -194,8 +198,9 @@ func SetBackend(b Backend) {
 }
 
 type defaultBackend struct {
-	HTTPClient *http.Client
-	URL        string
+	HTTPClient      *http.Client
+	URL             string
+	UserAgentSuffix *string
 }
 
 // Call sends requests to the Clerk API and handles the responses.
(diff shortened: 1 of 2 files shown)


clerk.go
```diff
@@ -219,7 +224,11 @@ func (b *defaultBackend) newRequest(ctx context.Context, apiReq *APIRequest) (*h
 	}
 	req.Header.Add("Authorization", fmt.Sprintf("Bearer %s", secretKey))
 	req.Header.Add("Content-Type", "application/json")
-	req.Header.Add("User-Agent", fmt.Sprintf("Clerk/%s SDK-Go/%s", clerkAPIVersion, sdkVersion))
+	userAgent := fmt.Sprintf("Clerk/%s SDK-Go/%s", clerkAPIVersion, sdkVersion)
```
georgepsarakis, line 227: ℹ️ Just mentioning that Cloudflare analytics group by user agent, but I guess we can always filter by prefix!

clerk_test.go
```diff
@@ -155,10 +158,10 @@ func TestBackendCall_RequestHeaders(t *testing.T) {
 		// The client sets the Authorization header correctly.
 		assert.Equal(t, fmt.Sprintf("Bearer %s", secretKey), r.Header.Get("Authorization"))
 		// The client sets the User-Agent header.
-		assert.Equal(t, "Clerk/v1 SDK-Go/v2.0.0", r.Header.Get("User-Agent"))
+		assert.Equal(t, fmt.Sprintf("Clerk/v1 SDK-Go/%s (clerk.com)", sdkVersion), r.Header.Get("User-Agent"))
```
georgepsarakis, line 161: 💡 Perhaps adding a custom string to aid in tracing the request here instead? Just to serve as a usage example.

georgepsarakis (approved):

clerk.go
gkats: Thanks, I didn't know that. Do you think it's better to drop this change. 

An alternative solution would be to leave the User-Agent string as is and maybe send a custom header instead. Would that be better?

kostaspt: @georgepsarakis Can't we just use the `X-Clerk-SDK` header?

clerk_test.go
georgepsarakis: Just a random-looking identifier maybe?

clerk.go
gkats: The `X-Clerk-SDK` is already used, but we could add another one for debugging purposes, or to uniquely identify a consumer, like one of our partners or us.

kostaspt: Yep, what I had in mind was to let the user set the User-Agent to whatever, not just the suffix, and rely on the `X-Clerk-SDK` for wherever we need to identify if it came from the SDK.

gkats: I see. That could work. However, I think we should have control over the User-Agent header, since it's a "standard" header but be loose in what we accept for custom ones.

gkats:
Closing this PR in favor of a solution which adds a custom header.

clerk.go
dimkl: @agis we are following this approach in for BAPI requests by setting the request user-agent to `{packageName}@{packageVersion}`.

ref: https://github.com/clerk/javascript/blob/main/packages/backend/src/constants.ts#L4

gkats: Here's an alternative, which uses a custom header. https://github.com/clerk/clerk-sdk-go/pull/252

PR #222, opened 2024-02-07
feat: Blocklist Identifiers API

gkats:
Added support for the blocklist identifiers API. Supported operations are Create, Delete and List.


actortoken/api.go
```diff
@@ -1,6 +1,6 @@
 // Code generated by "gen"; DO NOT EDIT.
 // This file is meant to be re-generated in place and/or deleted at any time.
-// Last generated at 2024-02-06 13:00:04.900202362 +0000 UTC
+// Last generated at 2024-02-07 11:36:32.720909014 +0000 UTC
```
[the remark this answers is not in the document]
georgepsarakis: I think it's fine to remove, although Github automatically hides the diff by default for auto-generated files.

blocklistidentifier/client.go
```diff
@@ -0,0 +1,67 @@
...
+type ListParams struct {
+	clerk.APIParams
+}
+
+// List returns all the identifiers in the blocklist.
+func (c *Client) List(ctx context.Context, params *ListParams) (*clerk.BlocklistIdentifierList, error) {
+	req := clerk.NewAPIRequest(http.MethodGet, fmt.Sprintf("%s?paginated=true", path))
+	list := &clerk.BlocklistIdentifierList{}
+	err := c.Backend.Call(ctx, req, list)
+	return list, err
```
georgepsarakis, line 66: ❓ Just a reminder that we don't return `nil` here, not sure where we landed on this!

georgepsarakis (approved):

blocklistidentifier/client.go
gkats: I don't think we landed anywhere, but it's possible that I might have missed us making a decision. 

Personally, I think that any return value cannot be trusted in case there's an error, so I'm fine with anything.

Let me know if you'd prefer changing the return type to always be `nil` when there's an error and I can do it.

georgepsarakis: All good, no objection, it's true that values must be discarded. Only benefit I could see is that we could avoid using pointers overall.

PR #223, opened 2024-02-07
feat: Clients API

gkats:
Added operations for the Clients API; Get, Verify and List. 

Added types for Client and Session.


clerktest/clerktest.go
```diff
@@ -34,11 +37,14 @@ func (rt *RoundTripper) RoundTrip(r *http.Request) (*http.Response, error) {
...
+	if rt.Method != "" {
+		require.Equal(rt.T, rt.Method, r.Method)
+	}
 	if rt.Path != "" {
 		require.Equal(rt.T, rt.Path, r.URL.Path)
 	}
-	if rt.Method != "" {
-		require.Equal(rt.T, rt.Method, r.Method)
+	if rt.Query != nil {
+		require.Equal(rt.T, rt.Query.Encode(), r.URL.Query().Encode())
```
georgepsarakis, line 47: ❓ Is order deterministic when encoding multiple parameters? It might be, just pointing out in case it causes any flakiness whatsoever. I think assertion functions should be capable of comparing the map object though 🤔

georgepsarakis (approved):

clerktest/clerktest.go
gkats: I thought about flakiness too, but I opted to leave it like this until an actual problem arises.

PR #224, opened 2024-02-07
feat: Email Addresses API

gkats:
Added support for the Email Addresses API operations Create, Read, Update and Delete.


email_address.go
```diff
@@ -0,0 +1,29 @@
...
+	Strategy                        string          `json:"strategy"`
+	Attempts                        *int64          `json:"attempts"`
+	ExpireAt                        *int64          `json:"expire_at"`
+	VerifiedAtClient                string          `json:"verified_at_client,omitempty"`
+	Nonce                           *string         `json:"nonce,omitempty"`
+	ExternalVerificationRedirectURL *string         `json:"external_verification_redirect_url,omitempty"`
+	Error                           json.RawMessage `json:"error,omitempty"`
+}
+
+type LinkedIdentification struct {
```
gkats, line 26: I don't like this name, so if you have something better to suggest, please do!

clerk.go
```diff
@@ -422,3 +422,8 @@ func String(v string) *string {
 func Int64(v int64) *int64 {
 	return &v
 }
+
+// Bool returns a pointer to the provided bool value.
+func Bool(v bool) *bool {
```
georgepsarakis, line 427: 🔧 Suggestion since this function is exported:

```suggestion
func BoolPtr(v bool) *bool {
```

email_address.go
georgepsarakis: Sounds fine to me. I would maybe omit the `Identification` prefix from the struct field names.

emailaddress/client_test.go
```diff
@@ -0,0 +1,149 @@
...
+		"code":"create-error-code"
+	}],
+	"clerk_trace_id":"create-trace-id"
+}`),
+		},
+	}
+	client := NewClient(config)
+	_, err := client.Create(context.Background(), &CreateParams{})
+	require.Error(t, err)
+	apiErr, ok := err.(*clerk.APIErrorResponse)
```
georgepsarakis, line 59: 💡 Perhaps a helper is useful?

```go
apiErr, ok := clerk.AsErrorResponse(err)
```

georgepsarakis (approved):

emailaddress/client_test.go
gkats: Good idea. However,  I cannot see how this would help. 

I imagine that if we change the `APIErrorResponse` type, even if we had the helper, we would still have to change the code that uses the error response type. :thinking: 

Happy to add the helper in a separate PR.

email_address.go
gkats: Done!

emailaddress/client_test.go
georgepsarakis: It would mostly serve as a guide. Autocompletion may help you write the conversion faster too. Otherwise someone must probably remember or look up the error type? The rest of the code will remain the same, you're absolutely right!

PR #226, opened 2024-02-08
feat: Organizations API

gkats:
Added support for the Organizations API. The available operations are Create, Read, Update, Delete, List, DeleteLogo and UpdateMetadata.


organization.go
```diff
@@ -0,0 +1,33 @@
...
+	PublicMetadata          json.RawMessage `json:"public_metadata"`
+	PrivateMetadata         json.RawMessage `json:"private_metadata"`
+	CreatedBy               string          `json:"created_by"`
+	CreatedAt               int64           `json:"created_at"`
+	UpdatedAt               int64           `json:"updated_at"`
+
+	// LogoURL is the URL of the organization's logo.
+	//
+	// Deprecated: Use ImageURL instead.
+	LogoURL *string `json:"logo_url,omitempty"`
```
chanioxaris, line 26: ❓ As this is a deprecated property, i believe we should stop using on the new major version

chanioxaris (approved):

organization.go
gkats: I have the same opinion. Same for deprecated endpoints.

However, we had a discussion with @alex-ntousias and he suggested that we keep everything that we support in v1 so that we make the transition to v2 easier.

I guess for struct fields with a clear alternative things are easier. I'll drop the field.

PR #227, opened 2024-02-08
feat: Organization Invitations API

gkats:
Added support for the Create operation of the Organization Invitations API.


organizationinvitation/client.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+}
+
+type CreateParams struct {
+	clerk.APIParams
+	EmailAddress    *string          `json:"email_address,omitempty"`
+	Role            *string          `json:"role,omitempty"`
+	RedirectURL     *string          `json:"redirect_url,omitempty"`
+	InviterUserID   *string          `json:"inviter_user_id,omitempty"`
+	PublicMetadata  *json.RawMessage `json:"public_metadata,omitempty"`
```
georgepsarakis, line 37: 🔧 I don't think we should be using pointers to [`json.RawMessage`](https://pkg.go.dev/encoding/json#RawMessage) since it's an alias for `[]byte`. It's actually slightly problematic to pass pointers to slices as they can suffer unintended mutations.

georgepsarakis (approved):

PR #228, opened 2024-02-08
feat: Cache requests for JWKS on JWT verification

gkats:
The jwt.Verify method needs to fetch the JSON Web Key Set from the API in order to verify the session JWT's validity. 

The jwt.Verify method is used in the http.WithHeaderAuthorization middleware, which means that in an HTTP server context, the method will executed for every request.

We're adding a caching layer for the JWKS when we verify the session JWT. This way we can cache the JWKS response from the API for 1 hour.

### DIFF
--- a/jwt/jwt.go
+++ b/jwt/jwt.go
@@ -6,6 +6,7 @@ import (
 	"context"
 	"fmt"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/clerk/clerk-sdk-go/v2"
@@ -103,10 +104,9 @@ func Verify(ctx context.Context, params *VerifyParams) (*clerk.SessionClaims, er
 	return claims, nil
 }
 
+// Retrieve the JSON web key for the provided id from the set.
 func getJWK(ctx context.Context, kid string) (*clerk.JSONWebKey, error) {
-	// TODO Avoid multiple requests by caching results for the same
-	// instance.
-	jwks, err := jwks.Get(ctx, &jwks.GetParams{})
+	jwks, err := getJWKSWithCache(ctx)
 	if err != nil {
 		return nil, err
 	}
@@ -116,6 +116,44 @@ func getJWK(ctx context.Context, kid string) (*clerk.JSONWebKey, error) {
 		}
 	}
 	return nil, fmt.Errorf("no jwk key found for kid %s", kid)
+}
+
+// Returns the JSON web key set. Tries a cached value first, but if
+// there's no value or the entry has expired, it will fetch the set
+// from the API and cache the value.
+func getJWKSWithCache(ctx context.Context) (*clerk.JSONWebKeySet, error) {
+	const cacheKey = "/v1/jwks"
+	var jwks *clerk.JSONWebKeySet
+	var err error
+
+	// Try the cache first. Make sure we have a non-expired entry and
+	// that the value is a valid JWKS.
+	entry, ok := getCache().Get(cacheKey)
+	if ok && !entry.HasExpired() {
+		jwks, ok = entry.GetValue().(*clerk.JSONWebKeySet)
+		if !ok || jwks == nil || len(jwks.Keys) == 0 {
+			jwks, err = forceGetJWKS(ctx, cacheKey)
+			if err != nil {
+				return nil, err
+			}
+		}
+	} else {
+		jwks, err = forceGetJWKS(ctx, cacheKey)
+		if err != nil {
+			return nil, err
+		}
+	}
+	return jwks, err
+}
+
+// Fetches the JSON web key set from the API and caches it.
+func forceGetJWKS(ctx context.Context, cacheKey string) (*clerk.JSONWebKeySet, error) {
+	jwks, err := jwks.Get(ctx, &jwks.GetParams{})
+	if err != nil {
+		return nil, err
+	}
+	getCache().Set(cacheKey, jwks, time.Now().UTC().Add(time.Hour))
+	return jwks, nil
 }
 
 func isValidIssuer(iss string) bool {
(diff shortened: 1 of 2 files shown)


gkats:
Please note that I'm planning to look into ways of inserting a "clock" for replacing `time` calls, so that we can have more control over testing.

The main concern is to figure out an API for a clock that does not depend on external libraries. I have some ideas, but I consider them out of scope for this PR.

jwt/jwt.go
```diff
@@ -154,3 +192,66 @@ func Decode(_ context.Context, params *DecodeParams) (*clerk.Claims, error) {
 		Extra:  extraClaims,
 	}, nil
 }
+
+// Caching store.
+type cache struct {
+	mu      sync.RWMutex
+	entries map[string]*cacheEntry
```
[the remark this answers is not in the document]
gkats: Good idea. We can optimistically prune every time we force fetch. I can add this as an enhancement in a later PR.

jwt/jwt.go
```diff
@@ -118,6 +118,44 @@ func getJWK(ctx context.Context, kid string) (*clerk.JSONWebKey, error) {
 	return nil, fmt.Errorf("no jwk key found for kid %s", kid)
 }
 
+// Returns the JSON web key set. Tries a cached value first, but if
+// there's no value or the entry has expired, it will fetch the set
+// from the API and cache the value.
+func getJWKSWithCache(ctx context.Context) (*clerk.JSONWebKeySet, error) {
+	cacheKey := "/v1/jwks"
```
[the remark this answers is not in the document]
gkats: Hmmm.... I guess not until it's used some place else?

There's more that we can offer to avoid cache key collisions, but I think for now we're good.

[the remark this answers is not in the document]
kostaspt: Mostly because it's not a "variable" if it's not changing. But yep, minor nit.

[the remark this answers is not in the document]
georgepsarakis: You can always define it as a constant inside the function, it's a good middle ground I believe.

georgepsarakis, line 125: ❓ Don't you need a mutex here?

jwt/jwt.go
```diff
@@ -118,6 +118,44 @@ func getJWK(ctx context.Context, kid string) (*clerk.JSONWebKey, error) {
...
+	var jwks *clerk.JSONWebKeySet
+	var err error
+
+	// Try the cache first. Make sure we have a non-expired entry and
+	// that the value is a valid JWKS.
+	entry, ok := getCache().Get(cacheKey)
```
georgepsarakis, line 131: 🔧 I think the cache must have the KID (instance ID) as key (or part of it). You can then quickly lookup by KID since that's the requirement from line 114. This way you can also have a multi-instance scenario handled.

jwt/jwt.go
```diff
@@ -118,6 +118,44 @@ func getJWK(ctx context.Context, kid string) (*clerk.JSONWebKey, error) {
...
+	return jwks, err
+}
+
+// Fetches the JSON web key set from the API and caches it.
+func forceGetJWKS(ctx context.Context, cacheKey string) (*clerk.JSONWebKeySet, error) {
+	jwks, err := jwks.Get(ctx, &jwks.GetParams{})
+	if err != nil {
+		return nil, err
+	}
+	getCache().Set(cacheKey, jwks, time.Now().Add(time.Hour))
```
georgepsarakis, line 155: 🔧 A mutex write lock may be required here as well.

georgepsarakis, line 155: 🔧 Perhaps use UTC everywhere to prevent inconsistencies.
```suggestion
	getCache().Set(cacheKey, jwks, time.Now().UTC().Add(time.Hour))
```

jwt/jwt.go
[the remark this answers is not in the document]
georgepsarakis: As far as I understand we currently store always one entry here, isn't that right?

jwt/jwt.go
```diff
@@ -154,3 +192,66 @@ func Decode(_ context.Context, params *DecodeParams) (*clerk.Claims, error) {
...
+type cache struct {
+	mu      sync.RWMutex
+	entries map[string]*cacheEntry
+}
+
+// Get returns the cache entry for the provided key, if one exists.
+func (c *cache) Get(key string) (*cacheEntry, bool) {
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+	entry, ok := c.entries[key]
```
georgepsarakis, line 206: ❓ If the entry is expired, I think you can ignore it here and don't return it?

jwt/jwt.go
gkats: There's already a mutex when we read/write from the cache. I believe it should be enough?

jwt/jwt.go
gkats: Good thinking. I'm just a bit reluctant to change the existing behavior. :thinking: 

As far as I understand the cache stores the whole json web key set response from the server. It was like this before and I'm a bit hesitant to change it now.

Regarding multi-instance scenarios, indeed, caching per instance is not possible with this setup, but I don't think it's needed for now. 

I'll toy a bit with your proposal and open a separate PR if it doesn't break anything. Assuming that's alright, since you added a :wrench: comment. :smile:

jwt/jwt.go
gkats: Was debating this as well. The entry is still there technically, but I guess users of a cache would expect to not get back expired entries?

jwt/jwt.go
gkats: See other comment regarding protecting reads/writes with a mutex. Do you think that's not enough?

jwt/jwt.go
georgepsarakis: I think you're right. Since the JWKS is now cached it's not an issue if we perform multiple requests.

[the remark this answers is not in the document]
gkats: Done!

jwt/jwt.go
gkats: Done, although I believe in this case it would be redundant, we're just adding duration.

jwt/jwt.go
georgepsarakis: If it's going to be replaced, sounds good to me to not return it. My suggestion just transfers the concern to the cache layer. Not a big deal.

jwt/jwt.go
georgepsarakis: You're right!

georgepsarakis: It's unusual but some servers will use local timezones. In that case, on DST changes you might observe weird corner cases. UTC also makes local testing more consistent (e.g. CI vs local dev, or cross-timezone teams).

PR #231, opened 2024-02-08
feat: Organization Memberships API

gkats:
Added support for the Organization Memberships API. Available operations are Create, Update, Delete and List.


clerk.go
```diff
@@ -297,7 +297,7 @@ func setRequestQuery(req *http.Request, params Params) {
 	paramsQuery := params.ToQuery()
 	for k, values := range paramsQuery {
 		for _, v := range values {
-			q.Set(k, v)
+			q.Add(k, v)
```
gkats, line 300: @georgepsarakis I think you've voiced a concern about "array" parameters in the query string. Your instinct was right. We need to add values and not overwrite them in order to support arrays.

organizationmembership/client_test.go
```diff
@@ -0,0 +1,207 @@
...
+}],
+"total_count": 1
+}`,
+				id, organizationID, userID)),
+			Method: http.MethodGet,
+			Path:   "/v1/organizations/" + organizationID + "/memberships",
+			Query: &url.Values{
+				"limit":    []string{"1"},
+				"offset":   []string{"2"},
+				"role":     []string{"admin", "member"},
```
georgepsarakis, line 189: ❓ Ah here's a case for an array, right? https://github.com/clerk/clerk-sdk-go/pull/231/files#r1482950532

georgepsarakis (approved):

PR #238, opened 2024-02-09
feat: Add create session token given a template slug endpoint

soulcodex:
- Add clerk backend [api endpoint](https://clerk.com/docs/reference/backend-api/tag/Sessions#operation/CreateSessionTokenFromTemplate) implementation to create a session token given a jwt template slug

### DIFF
--- a/clerk/sessions.go
+++ b/clerk/sessions.go
@@ -18,6 +18,11 @@ type Session struct {
 	UpdatedAt                int64  `json:"updated_at"`
 }
 
+type SessionToken struct {
+	Object string `json:"object"`
+	JWT    string `json:"jwt"`
+}
+
 type ListAllSessionsParams struct {
 	Limit    *int
 	Offset   *int
@@ -112,4 +117,16 @@ func (s *SessionsService) Verify(sessionId, token string) (*Session, error) {
 		return nil, err
 	}
 	return &sessionResponse, nil
+}
+
+func (s *SessionsService) CreateTokenFromTemplate(sessionID, templateSlug string) (*SessionToken, error) {
+	sessionURL := fmt.Sprintf("%s/%s/token/%s", SessionsUrl, sessionID, templateSlug)
+	req, _ := s.client.NewRequest("POST", sessionURL)
+
+	var sessionToken SessionToken
+	_, err := s.client.Do(req, &sessionToken)
+	if err != nil {
+		return nil, err
+	}
+	return &sessionToken, nil
 }
--- a/clerk/sessions_test.go
+++ b/clerk/sessions_test.go
@@ -3,6 +3,8 @@ package clerk
 import (
 	"encoding/json"
 	"fmt"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	"net/http"
 	"net/url"
 	"reflect"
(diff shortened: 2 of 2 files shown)


tests/integration/sessions_test.go
```diff
@@ -27,5 +27,10 @@ func TestSessions(t *testing.T) {
 		if session == nil {
 			t.Fatalf("Sessions.Read returned nil")
 		}
+
+		token, tokenErr := client.Sessions().CreateTokenFromTemplate(sessionId, "someTemplateSlug")
+		if tokenErr != nil {
+			t.Fatalf("Sessions.CreateTokenFromTemplate returned tokenError: %v", err)
+		}
```
soulcodex, line 34: IDK if this is going to work as expected given that the token slug may not exist at the clerk account used to run these E2E :sweat_smile:

clerk/sessions.go
```diff
@@ -18,6 +18,11 @@ type Session struct {
 	UpdatedAt                int64  `json:"updated_at"`
 }
 
+type SessionToken struct {
+	Object string `json:"object"`
+	Jwt    string `json:"jwt"`
```
gkats, line 23: Let's rename this to `JWT` 

https://go.dev/talks/2014/names.slide#5

clerk/sessions.go
```diff
@@ -113,3 +118,15 @@ func (s *SessionsService) Verify(sessionId, token string) (*Session, error) {
 	}
 	return &sessionResponse, nil
 }
+
+func (s *SessionsService) CreateTokenFromTemplate(sessionId, templateSlug string) (*SessionToken, error) {
+	sessionUrl := fmt.Sprintf("%s/%s/token/%s", SessionsUrl, sessionId, templateSlug)
```
gkats, line 123: Same here for renaming to  `sessionURL` and `sessionID`. :wink:

https://go.dev/talks/2014/names.slide#5

tests/integration/sessions_test.go
gkats: Let's remove the integration test completely. 

You did the right thing to add it here, but your instinct is correct. It won't work unless the template slug exists on the instance.

clerk/sessions_test.go
```diff
@@ -242,6 +243,47 @@ func TestSessionsService_Verify_invalidServer(t *testing.T) {
 	}
 }
 
+func TestSessionsService_CreateTokenFromTemplate_Success(t *testing.T) {
+	token := "token"
+	sessionId := fmt.Sprintf("sess_%s", "2bp7pawkvFUR3m3QPBm6Cwx5ghZ")
```
gkats, line 248: This `Sprintf` call is probably not needed. There's no variable substitution here. Same for line 278 below.

clerk/sessions_test.go
```diff
@@ -242,6 +243,47 @@ func TestSessionsService_Verify_invalidServer(t *testing.T) {
...
+	assert.NoError(t, err)
+
+	received, err := clerkClient.Sessions().CreateTokenFromTemplate(sessionId, templateSlug)
+	equal := reflect.DeepEqual(*received, expectedSessionToken)
+	assert.NoError(t, err)
+	assert.Truef(
+		t,
+		equal,
+		"Response = %v, want %v", *received, expectedSessionToken,
+	)
```
gkats, line 274: I believe we can assert that the response is correctly parsed as well. How's this for an assertion?

```go
got, err := clerkClient.Sessions().CreateTokenFromTemplate(sessionId, templateSlug)
require.NoError(t, err)
require.Equal(t, "token", got.Object)
require.Equal(t, sessionToken, got.JWT)
```

clerk/sessions_test.go
```diff
@@ -242,6 +243,47 @@ func TestSessionsService_Verify_invalidServer(t *testing.T) {
...
+}
+
+func TestSessionsService_CreateTokenFromTemplate_InvalidServer(t *testing.T) {
```
gkats, line 277: Feel free to skip/remove this test.

clerk/sessions_test.go
```diff
@@ -252,3 +294,10 @@ const dummySessionJson = `{
         "status": "ended",
         "user_id": "user_1mebQggrD3xO5JfuHk7clQ94ysA"
     }`
+
+const dummySessionTokenJsonTemplate = `
```
gkats, line 298:
```suggestion
const dummySessionTokenJSONTemplate = `
```

Looks like this constant is used only in one place. We can add it inside the test where it's actually used.

soulcodex:
> Thank you for your contribution @soulcodex!! 😍
> 
> Overall looks great to me. I've left some comments regarding variable naming and an alternative assertion for the tests.
> 
> Please remove the integration test, it depends on the instance setup and cannot be trusted. It's not your fault, but ours. 😄
> 
> Out of curiosity, what led you to add support in the SDK for this endpoint? Do you intend to use it?

The changes requested has been made :wink:

The main motivation for this proposal is because I'm trying to run on my side project an acceptance / E2E tests suite in order to verify that my securized endpoints are working as expected passing against every single layer to produce the expected result / behavior :smile: 

image

Unfortunately to be able to check E2E my backend I need to issue a session using the Clerk FAPI and then request a JWT token given an slug (JWT Template) using the Clerk BAPI SDK to finally store the JWT token in a Redis during a certain period of time (TTL) to avoid reach the api "rate limiter" issuing the token on each scenario / iteration of the test suite :disappointed: :roller_coaster: 

However feel free to request any other change but this is basically my adventure / motivation about this PR :sweat_smile:

clerk/sessions_test.go
gkats: You are right! :smile:

gkats:
> However feel free to request any other change but this is basically my adventure / motivation about this PR 😅

Thanks for providing such detailed information! I just wanted to get a sense of real-world usage patterns.

Thank you for your contribution!

gkats (approved):

PR #239, opened 2024-02-10
feat: Instance Settings API

gkats:
API operations for instance settings. Update instance settings, update instance restrictions and update instance organization settings.


clerk_test.go
```diff
@@ -279,6 +279,35 @@ func TestBackendCall_SuccessfulResponse_GetRequest(t *testing.T) {
 	assert.JSONEq(t, rawJSON, string(resource.Response.RawJSON))
 }
 
+// TestBackendCall_SuccessfulResponse_EmptyResponse tests successful
+// responses with no response body.
+func TestBackendCall_SuccessfulResponse_EmptyResponse(t *testing.T) {
```
georgepsarakis, line 284: ➕

cmd/gen/main.go
```diff
@@ -129,8 +129,8 @@ func main() {
 const lineStartsWith = "func (c *Client) "
 
 var nameRE = regexp.MustCompile("^\\w+\\(")
-var argsRE = regexp.MustCompile("^.+\\)\\s\\(")
-var returnRE = regexp.MustCompile("^.+\\)")
+var argsRE = regexp.MustCompile("^.+\\)\\s[\\(e]")
+var returnRE = regexp.MustCompile("^.+\\s{")
```
georgepsarakis, line 133: 💡 Mostly an idea/suggestion for possible future improvement, definitely not a current action item: we can use AST parsing to perform code generation in a possibly more maintainable way in the future: https://eli.thegreenplace.net/2021/rewriting-go-source-code-with-ast-tooling/

georgepsarakis (approved):

cmd/gen/main.go
gkats: Yes, this "generator" is a disgrace. :disappointed:

georgepsarakis: Gets the work done is what matters 🥇

PR #242, opened 2024-02-13
feat: Cache per JWKS client

gkats:
Moved the previously unexported cache for jwks keys in the jwks package and exported an interface that can be used to get and set values from and to the cache.

Added a jwks.Client that can invoke the JWKS endpoint. Accepts a clerk.Cache.

The whole point of having a Cache per jwks.Client is cases where the default singleton cache might cause problems with shared resource access.

For users that configure the SDK to work with a single Clerk API secret key, the default cache will do just fine. However, if you're planning to use the API clients with different keys, we should make sure that each client writes to a different cache.

We add the Cache to the client and provide an option to pass an existing client to the http.WithHeaderAuthorization middleware.


clerk.go
```diff
@@ -388,12 +389,12 @@ func (params ListParams) ToQuery() url.Values {
...
 }
 
-// Regular expression that matches multiple backslashes in a row.
-var extraBackslashesRE = regexp.MustCompile("([^:])//+")
-
 // JoinPath returns a URL string with the provided path elements joined
 // with the base path.
 func JoinPath(base string, elem ...string) (string, error) {
+	// Regular expression that matches multiple backslashes in a row.
+	var extraBackslashesRE = regexp.MustCompile("([^:])//+")
```
georgepsarakis, line 396: ❓ This change will cause the regular expression to recompile, with a slight overhead. I think the previous version was fine, did you find a problem with it?

jwks/client.go
```diff
@@ -0,0 +1,127 @@
...
+// Cache provides a store for JWKS values.
+type Cache interface {
+	// Get retrieves an entry from the cache.
+	Get() *clerk.JSONWebKeySet
+	// Set sets an entry in the cache.
+	Set(*clerk.JSONWebKeySet, time.Time)
+}
+
+// Caching store for JSON Web Key Sets.
+type defaultCache struct {
```
georgepsarakis, line 87: 🔧 I would suggest renaming the struct to `jwksCache` and use `defaultCache` as the global variable name in line 116.

jwt/jwt.go
```diff
@@ -147,12 +146,12 @@ func getJWKSWithCache(ctx context.Context) (*clerk.JSONWebKeySet, error) {
 }
 
 // Fetches the JSON web key set from the API and caches it.
-func forceGetJWKS(ctx context.Context, cacheKey string) (*clerk.JSONWebKeySet, error) {
-	jwks, err := jwks.Get(ctx, &jwks.GetParams{})
+func forceGetJWKS(ctx context.Context, client *jwks.Client) (*clerk.JSONWebKeySet, error) {
+	jwks, err := client.Get(ctx, &jwks.GetParams{})
```
georgepsarakis, line 150: 💡 Not sure, but I would expect that if a Client has caching enabled, then calls to `Get` can either fetch a cached entry or perform an HTTP request and add the cache entry. So in a sense this concern could be transferred to the client itself. I believe that could make the Client more directly usable for other usages, but I may be missing something.

georgepsarakis (commented):
Just added some suggestions, let me know what you think!

clerk.go
gkats: No, I was experimenting a bit in a WIP branch and messed it up. Thanks for noticing.

jwt/jwt.go
gkats: You're probably right. I also thought about this, but then I thought, what if we add more methods to the jwks.Client? Would we expect every fetching method to cache results? Would it make sense to name the cache `GetCache` to make it more obvious?

It feels weird to attach a cache and leave it to the responsibility of the consumer how it will be used.

On the other hand, caching is used only for `jwt.Verify`, so it makes sense that this method should be responsible to use the cache as it sees fit. Making decisions about when to cache, what to cache, how long until the cache expires, etc...

Also, it might make more sense to keep the Client.Get operation free of any caching. You can use the cache when you want, but you can also force-fetch.

I was conflicted to be honest. Perhaps your comment enforces the idea that "a jwks.Client with a cache will only fetch after 1h" and that's it?

georgepsarakis: > Would it make sense to name the cache GetCache to make it more obvious?

I think renaming the setting to `EnableReadCache` or `EnableGetCache` if you want to be more exact, sounds fine to me. If I'm thinking this right, you then might be able to hide the `Cache` interface entirely, so it becomes completely internal.

> It feels weird to attach a cache and leave it to the responsibility of the consumer how it will be used.

That's my intuition too. It feels that the storage layer is part of the client but the caller must perform all interactions with the caching layer.

> You can use the cache when you want, but you can also force-fetch.

Wouldn't you just create a new Client instance then? What's the scenario where you need to force fetch considering that you've made the decision to enable caching? Either way, you can maybe expose a `ResetCache` method.

gkats:
Closing in favor of #244

PR #244, opened 2024-02-13
feat: JWKS cache is the HTTP middleware's concern

gkats:
The JSON Web Key Set for an instance is not likely to change often. We're taking advantage of this fact and save requests to the Clerk API when verifying a JWT, by caching the API's GET /v1/jwks responses. 

In reality, caching is needed by the HTTP middleware. We don't want to fetch the JWKS for every request that the middleware needs to handle. This commit moves the caching layer to the http middleware. 

The cache is repurposed to store JSON Web Keys under their respective KeyIDs. The jwt.Verify function now requires the JSON Web Key to be present in the parameters and it is the responsibility of the caller to retrieve it.

This way we achieve a clearer separation of concerns. The jwt.Verify function should not be responsible for fetching the JWK, nor cache it. It's the http middleware that needs the caching layer.


http/middleware.go
```diff
@@ -152,3 +235,72 @@ func JSONWebKey(key string) AuthorizationOption {
...
+var cacheInit sync.Once
+
+// A "singleton" JWK cache for the package.
+var cache *jwkCache
+
+// GetCache returns the library's default cache singleton.
+// Please note that the returned Cache is a package-level variable.
+// Using the package with more than one Clerk API secret keys might
+// require to use different Clients with their own Cache.
+func GetCache() *jwkCache {
```
georgepsarakis, line 299: :question: Do we need these functions to be exported? Maybe for testing?

georgepsarakis (approved):

http/middleware.go
gkats: We don't. Updated.

http/middleware.go
```diff
@@ -32,8 +37,22 @@ func RequireHeaderAuthorization(opts ...AuthorizationOption) func(http.Handler)
...
 // Authorization: Bearer <token>
 func WithHeaderAuthorization(opts ...AuthorizationOption) func(http.Handler) http.Handler {
+	var paramsErr error
+	params := &AuthorizationParams{}
+	for _, opt := range opts {
+		paramsErr = opt(params)
+		if paramsErr != nil {
+			break
+		}
+	}
```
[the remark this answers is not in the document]
georgepsarakis: I agree!

PR #247, opened 2024-02-14
feat: Handler to check azp claim

gkats:
The JWT verification function accepts a handler that can be used to validate the 'azp' claim of the token.

The WithHeaderAuthorization middleware accepts either a handler or a list of authorized parties that must contain the 'azp' claim.

You can use the option in two ways.

```go
func checkAZP(azp string) bool {
  // do whatever you want with the azp claim
}

WithHeaderAuthorization(AuthorizedParty(checkAZP))
// OR
// This will check that the 'azp' claim is one of the provided values
WithHeaderAuthorization(AuthorizedPartyFunc("clerk.com", "clerk.dev"))
```

### DIFF
--- a/http/middleware.go
+++ b/http/middleware.go
@@ -160,12 +160,31 @@ type AuthorizationParams struct {
 // authorization options.
 type AuthorizationOption func(*AuthorizationParams) error
 
-// AuthorizedParty sets the authorized parties that will be checked
-// against the azp JWT claim.
-func AuthorizedParty(parties ...string) AuthorizationOption {
+// AuthorizedParty allows to provide a handler that accepts the
+// 'azp' claim.
+// The handler can be used to perform validations on the azp claim
+// and should return false to indicate that something is wrong.
+func AuthorizedParty(handler func(string) bool) AuthorizationOption {
 	return func(params *AuthorizationParams) error {
-		params.SetAuthorizedParties(parties...)
+		params.AuthorizedPartyHandler = handler
 		return nil
+	}
+}
+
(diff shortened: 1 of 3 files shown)


http/middleware.go
```diff
@@ -160,15 +160,34 @@ type AuthorizationParams struct {
...
 	return func(params *AuthorizationParams) error {
-		params.SetAuthorizedParties(parties...)
+		params.AuthorizedPartyFunc = handler
 		return nil
 	}
 }
 
+// AuthorizedPartyFunc registers a handler that checks that the
+// 'azp' claim's value is included in the provided parties.
+func AuthorizedPartyFunc(parties ...string) func(string) bool {
```
yourtallness, line 176: As discussed in Slack, this personally confuses me.

Maybe a naming that conveys what it does (check containment in list) instead of a name that coincides with the param below.

gkats: @yourtallness I made a couple of changes in https://github.com/clerk/clerk-sdk-go/pull/247/commits/a23a3917a1afc471ddd169b0c6a19bce843f3eb4

- Renamed the verification parameter from AuthorizedPartyFunc to AuthorizedPartyHandler
- Renamed the option helper in the middleware from AuthorizedPartyFunc to AuthorizedPartyMatches

Please let me know what you think, thanks!

yourtallness: This is an improvement.

The ideal UX would probably automatically call the default matcher if no handler is provided, but I understand this may be impractical.

So, approving AFAIAC.

yourtallness (approved):

PR #248, opened 2024-02-15
feat: Multipart requests

gkats:
Added support for the organization.UpdateLogo function, which can perform multipart/form-data requests.

Enhanced our default Backend's Call method so that it now supports multipart request bodies.

The Params interface gets one more method, ToMultipart(), so that any type that will be used as a multipart request body can transform itself to an appropriate multipart message.


clerk.go
```diff
@@ -267,19 +286,32 @@ func (b *defaultBackend) do(req *http.Request, params Params, setter ResponseRea
...
-		return err
+	var body []byte
+	var err error
+	if apiReq.isMultipart {
+		var contentType string
+		body, contentType, err = apiReq.Params.ToMultipart()
+		if err != nil {
+			return err
+		}
+		req.Header.Set("Content-Type", contentType)
```
georgepsarakis, line 309: ❓ Is the content type dynamic here or is it a specific value always?

clerk_test.go
```diff
@@ -385,3 +412,38 @@ func TestBackendCall_NonParseableError(t *testing.T) {
...
+
+		_, err = w.Write([]byte(`{}`))
+		require.NoError(t, err)
+	}))
+	defer ts.Close()
+
+	// Set up a mock backend which triggers requests to our test server above.
+	SetBackend(NewBackend(&BackendConfig{
+		HTTPClient: ts.Client(),
+		URL:        &ts.URL,
```
georgepsarakis, line 437: 🙃 I just noticed this now: should this field be names `URL` or `BaseURL` 🤔 ?

organization/client.go
```diff
@@ -117,6 +120,65 @@ func (c *Client) Delete(ctx context.Context, id string) (*clerk.DeletedResource,
 	return organization, err
 }
 
+type UpdateLogoParams struct {
+	clerk.APIParams
+	File           multipart.File `json:"-"`
+	UploaderUserID *string        `json:"-"`
+	Filename       *string        `json:"-"`
```
georgepsarakis, line 127: ❓ Is the filename parameter useful? Do we use it for mimetype detection?

georgepsarakis (approved):

chanioxaris (approved):

clerk_test.go
gkats: I think I like `URL`.

clerk.go
gkats: The content-type is multipart/form-data but also contains the writer's boundary.

https://cs.opensource.google/go/go/+/refs/tags/go1.22.0:src/mime/multipart/writer.go;l=74

I believe this is how the server knows how to parse the body successfully and figure out where each part ends.

https://datatracker.ietf.org/doc/html/rfc7578#section-4.1

organization/client.go
gkats: Looks like it's not useful at all. Removed. :smile:

clerk.go
georgepsarakis: TIL, thanks for explaining!

PR #250, opened 2024-02-15
feat: Upgrade guide

gkats:
Added upgrade guide and changelog templates for the library. 

Added an entry in the upgrade guide for migrating from v1 to v2.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,6 @@
+# Changelog
+
+## 2.0.0
+
+- Initial version for changelog.
+- Complete rewrite, new library API.
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -1,0 +1,372 @@
+# Clerk Go SDK upgrade guide
+
+The Clerk Go SDK uses [Semantic Versioning](https://semver.org/)
+for tracking all changes to the library.
+
+Semantic versions take the form of `MAJOR.MINOR.PATCH`. Incrementing
+
+- MAJOR version introduces incompatible API changes
+- MINOR version introduces functionality in a backwards compatible manner
+- PATCH version includes backwards compatible bug fixes
+
+This file contains upgrade notes for all `MAJOR` version bumps.
+
+While there are no plans to sunset older versions of the Go SDK,
+the current `MAJOR` version of `clerk-sdk-go` is actively supported.
+
+Bug fixes and security updates may be added to previous versions, but
+there is no guarantee.
+
+## 1.x.x to 2.x.x
+
+The `v2` version of the Clerk Go SDK is a complete rewrite and introduces
+a lot of breaking changes.
+
+### Minimum Go version
+
+The minimum supported Go version for the `v2` version of the Clerk Go SDK is `1.19`.
+
+### New API and package layout
+
+API operations in `v1` of the Clerk Go SDK are organized by service. There are
+different services for each API and all services are properties of a single
+`clerk.Client`.
+
+Let's see a concrete example. Assume that we want to create an organization
+and then list all available organizations.
+
+Here's how we would do it in `v1`.
+
+```go
+import (
+    "github.com/clerkinc/clerk-sdk-go"
+)
+
+
+func main() {
+    client, err := clerk.NewClient("sk_live_XXX")
+    if err != nil {
+        // handle error
+    }
+
+    // Create an organization
+    org, err := client.Organizations().Create(clerk.CreateOrganizationParams{
+        Name: "Acme Inc",
+    })
+    if err != nil {
+        if errResp, ok := err.(*clerk.ErrorResponse); ok {
+            // Access the API errors
+            errResp.Errors
+        }
+    }
+    // List all organizations, limit results to one.
+    limit := 1
+    orgs, err := client.Organizations().ListAll(clerk.ListAllOrganizationsParams{
+        Limit: &limit,
+    })
+    if err != nil {
+        // handle the error
+    }
+    if orgs.TotalCount > 0 {
+        // Get the first organization in the list
+        org = orgs.Data[0]
+    }
+}
+```
+
+The `v2` version breaks away of the client and services pattern seen above.
+In `v2`, every API has its own package, following a resource-based structure.
+
+There are two ways to call API operations in `v2`: with or without a client. Let's see both approaches.
+
+### Usage without a client
+
+In most cases, you'll only have to deal with a single API key in your project.
+
+```go
+import (
+    "context"
+
+    "github.com/clerk/clerk-sdk-go/v2"
+    "github.com/clerk/clerk-sdk-go/v2/organization"
+)
+
+func main() {
+    ctx := context.Background()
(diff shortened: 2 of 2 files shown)


UPGRADING.md
```diff
@@ -0,0 +1,317 @@
...
+    "github.com/clerkinc/clerk-sdk-go"
+)
+
+client, err := clerk.NewClient("sk_live_XXX")
+if err != nil {
+    // handle error
+}
+
+// Create an organization
+org, err := clerk.Organizations().Create(clerk.CreateOrganizationParams{
```
chanioxaris, line 51: This example should use the client variable right?
```suggestion
org, err := client.Organizations().Create(clerk.CreateOrganizationParams{
```

UPGRADING.md
```diff
@@ -0,0 +1,317 @@
...
+})
+if err != nil {
+    if errResp, ok := err.(*clerk.ErrorResponse); ok {
+        // Access the API errors
+        errResp.Errors
+    }
+}
+// List all organizations, limit results to one.
+limit := 1
+orgs, err := clerk.Organizations().ListAll(clerk.ListAllOrganizationsParams{
```
chanioxaris, line 62: Same here
```suggestion
orgs, err := client.Organizations().ListAll(clerk.ListAllOrganizationsParams{
```

UPGRADING.md
gkats: Indeed. Fixed.

UPGRADING.md
gkats: Updated.

UPGRADING.md
```diff
@@ -0,0 +1,321 @@
...
+## 1.x.x to 2.x.x
+
+The `v2` version of the Clerk Go SDK is a complete rewrite and introduces
+a lot of breaking changes.
+
+### Minimum Go version
+
+The minimum supported Go version for the `v2` version of the Clerk Go SDK is `1.19`.
+
+### Completely different API
```
georgepsarakis, line 29: ❓ Not sure exactly what alternative to suggest but just noting in case you can think of better phrasing to describe this section 🤔

UPGRADING.md
```diff
@@ -0,0 +1,321 @@
...
+    // handle the error
+}
+if orgs.TotalCount > 0 {
+    // Get the first organization in the list
+    org = orgs.Data[0]
+}
+```
+
+The `v2` version breaks away of the client and services pattern seen above.
+In `v2`, every API has its own package, following a resource-based organization.
```
georgepsarakis, line 75: ❓ Is it possible to avoid the term `organization` here? Structure maybe? Just noting because we're interacting with the Organization endpoints above!

UPGRADING.md
```diff
@@ -0,0 +1,321 @@
...
+
+In most cases, you'll only have to deal with a single API key in your project.
+
+```go
+import (
+    "github.com/clerk/clerk-sdk-go/v2"
+    "github.com/clerk/clerk-sdk-go/v2/organization"
+)
+
+ctx := context.Background()
```
georgepsarakis, line 89: 🔧 Perhaps we can do a few changes to convert this to a valid `main` script?

gkats: Done.

UPGRADING.md
gkats: Good point. Updated.

UPGRADING.md
georgepsarakis: Perhaps you or others can come up with something better but somethine like `New API and Package Layout` (or `Structure`)?

gkats: Sounds good to me. Updated. :smile:

CHANGELOG.md
```diff
@@ -0,0 +1,6 @@
+# Changelog
+
+## 2.0.0
+
+- Initial version for changelog
+- Complete rewrite, new APIs.
```
georgepsarakis, line 6: ❓ Any chance someone might be confused that new endpoints are supported?

UPGRADING.md
```diff
@@ -0,0 +1,372 @@
...
+with Clerk; `WithHeaderAuthorization` and `RequireHeaderAuthorization`.
+
+As the name implies, the new middleware support only authentication with a bearer token, that needs to be present in the
+"Authorization" header of the HTTP request.
+
+Cookie based authentication is not supported at all by the `v2` version of the library.
+
+Usage has also changed between `v1` and `v2`, as have the different options that the middleware support.
+
+Here's how you would use the `WithSessionV2` middleware to get access to the session token claims in HTTP handlers.
```
georgepsarakis, line 244: ❓ Maybe we can make it a bit more prominent that this sample refers to V1?

UPGRADING.md
```diff
@@ -0,0 +1,372 @@
...
+    "github.com/clerkinc/clerk-sdk-go"
+)
+
+func main() {
+    client, err := clerk.NewClient("sk_live_XXX")
+    if err != nil {
+        panic(err)
+    }
+    mux := http.NewServeMux()
+    mux.Handle("/session", clerk.WithSessionV2(client)(http.HandlerFunc(handleSession)))
```
georgepsarakis, line 259: 🔧 Perhaps use the `Require*` variants in all examples? I understand verification should be the default, unless there's some other use case?

UPGRADING.md
```diff
@@ -0,0 +1,372 @@
...
+    clerkhttp "github.com/clerk/clerk-sdk-go/v2/http"
+    "github.com/clerk/clerk-sdk-go/v2/jwks"
+)
+
+func main() {
+    config := &jwks.ClientConfig{}
+    config.Key = "sk_live_XXX"
+    client := jwks.NewClient(config)
+    mux := http.NewServeMux()
+    withAuth := clerkhttp.WithHeaderAuthorization(
```
georgepsarakis, line 317: 👍

UPGRADING.md
```diff
@@ -0,0 +1,372 @@
...
+
+The method accepts the same parameters, with two important differences.
+
+- The JSON web key with which the token will be verified is a required parameter.
+- The method will not cache the JSON web key.
+
+In the `v1` version, the `clerk.VerifyToken` method would trigger an HTTP request to the Clerk Backend API to
+fetch the JSON web key and would cache its value for one hour.
+
+However, fetching the key and caching it should not be the responsibility of the verification process so low
```
georgepsarakis, line 360: ❓ Not sure if the motivation is necessary, but I believe it may be welcomed. You can consider linking an example to how the middleware performs the caching process so that:
- There's no misunderstanding that the middleware moves away from caching.
- Someone can see an actual real-world example.

georgepsarakis (approved):

UPGRADING.md
georgepsarakis: Yes, but that seems as a design choice. Shouldn't we recommend the most secure integration?

UPGRADING.md
gkats: For sure. I'll change this whole section in a follow-up PR.

PR #251, opened 2024-02-15
fix: Change User-Agent header

gkats:
Changed the User-Agent header to match the rest of our SDKs. The pattern is package@version.


georgepsarakis (approved):
Nice catch!

PR #252, opened 2024-02-15
feat: Custom request header with application info

gkats:
Added the ability to provide application-specific info in API request headers.

The X-Clerk-Application header can contain a unique identifier for an application, such as its domain or ID.

Use of this header will allow us to quickly filter requests, debug and resolve support issues easier.

### DIFF
--- a/clerk.go
+++ b/clerk.go
@@ -157,6 +157,20 @@ type Params interface {
 	ToMultipart() ([]byte, string, error)
 }
 
+// CustomRequestHeaders contains predefined values that can be
+// used as custom headers in Backend API requests.
+type CustomRequestHeaders struct {
+	Application string
+}
+
+// Apply the custom headers to the HTTP request.
+func (h *CustomRequestHeaders) apply(req *http.Request) {
+	if h == nil {
+		return
+	}
+	req.Header.Add("X-Clerk-Application", h.Application)
+}
+
 // BackendConfig is used to configure a new Clerk Backend.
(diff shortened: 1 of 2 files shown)


clerk.go
```diff
@@ -169,6 +169,11 @@ type BackendConfig struct {
 	// Key is the Clerk secret key. If it's not set, the package level
 	// secretKey will be used.
 	Key *string
+	// Application allows you to provide information for your application
+	// that will be set in API request headers. Good candidates are the
+	// application's domain or ID, but possible values are not limited to
+	// those.
+	Application *string
```
georgepsarakis, line 176: ❓ I think it might be best to nest this option under a `DefaultHeaders` struct?

gkats: You mean something like this, or a `map[string][]string`?

```go
type BackendConfig struct {
  DefaultHeaders Headers 
}

type Headers struct {
  Application string
}
```

I guess you're thinking that we might want to be able to add more custom headers in the future right?

georgepsarakis: I was thinking the same exact structure as in your snippet (and avoid a `map`). Struct fields can be more informative/self-documented in a way. 
I think there's already a project that may introduce a default request header (not sure if we plan to provide it through this struct though but could be an option).

gkats: Updated. Please take another look, thanks!

georgepsarakis (approved):
💯

PR #253, opened 2024-02-16
feat: JWT parsing with custom claims

gkats:
Replaced the CustomClaims parameter with a CustomClaimsConstructor function when verifying a session JWT. The option is also available in the HTTP middleware.

The constructor function will be called when the JWT is parsed, producing a new struct instance instead of writing on a single instance. 

The custom claims will be made available in the SessionClaims.Custom field.

### DIFF
--- a/http/middleware.go
+++ b/http/middleware.go
@@ -188,11 +188,30 @@ func AuthorizedPartyMatches(parties ...string) func(string) bool {
 	}
 }
 
-// CustomClaims allows to pass a type (e.g. struct), which will be populated with the token claims based on json tags.
-// You must pass a pointer for this option to work.
-func CustomClaims(claims any) AuthorizationOption {
+// CustomClaimsConstructor allows to pass a constructor function
+// which returns a pointer to a type (struct) to hold custom token
+// claims.
+// The instance of the custom claims type will be then made available
+// through the clerk.SessionClaims struct.
+//
+//	// Define a type to describe the custom claims.
+//	type MyCustomClaims struct {
+//		ACustomClaim string `json:"a_custom_claim"`
+//	}
+//
+//	// In your HTTP server mux, configure the middleware with
+//	// the custom claims constructor.
+//	WithHeaderAuthorization(CustomClaimsConstructor(func(_ context.Context) any {
+//		return &MyCustomClaims{}
+//	})
+//
+//	// In the HTTP handler, access the active session claims. The
+//	// custom claims are available in the SessionClaims.Custom field.
+//	sessionClaims, ok := clerk.SessionClaimsFromContext(r.Context())
+//	customClaims, ok := sessionClaims.Custom.(*MyCustomClaims)
+func CustomClaimsConstructor(constructor func(context.Context) any) AuthorizationOption {
 	return func(params *AuthorizationParams) error {
-		params.CustomClaims = claims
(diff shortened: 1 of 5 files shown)


http/middleware.go
```diff
@@ -188,11 +188,30 @@ func AuthorizedPartyMatches(parties ...string) func(string) bool {
...
+//	// the custom claims constructor.
+//	WithHeaderAuthorization(CustomClaimsConstructor(func() any {
+//		return &MyCustomClaims{}
+//	})
+//
+//	// In the HTTP handler, access the active session claims. The
+//	// custom claims are available in the SessionClaims.Custom field.
+//	sessionClaims, ok := clerk.SessionClaimsFromContext(r.Context())
+//	customClaims, ok := sessionClaims.Custom.(*MyCustomClaims)
+func CustomClaimsConstructor(constructor func() any) AuthorizationOption {
```
georgepsarakis, line 212: ❓ Would it make sense to pass the request context in the constructor?

```suggestion
func CustomClaimsConstructor(constructor func(ctx context.Context) any) AuthorizationOption {
```

jwt/jwt.go
```diff
@@ -12,15 +12,32 @@ import (
...
 	Token string
 	// JWK the custom JSON Web Key that will be used to verify the
 	// Token with. Required.
-	JWK          *clerk.JSONWebKey
-	CustomClaims any
+	JWK *clerk.JSONWebKey
+	// CustomClaimsConstructor will be called when parsing the Token's
+	// claims. It's useful for parsing custom claims into user-defined
+	// types.
+	// Make sure it returns a pointer to a type (struct) that describes
```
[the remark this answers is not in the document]
georgepsarakis: On second thought, it might be clearer to always request a pointer, just trying to think how the failure is manifested if a value is returned.

jwt/jwt.go
```diff
@@ -80,7 +94,10 @@ func Verify(ctx context.Context, params *VerifyParams) (*clerk.SessionClaims, er
 	return claims, nil
 }
 
-func isValidIssuer(iss string) bool {
+func isValidIssuer(iss string, proxyURL *string) bool {
+	if proxyURL != nil {
```
georgepsarakis, line 98: ❓ I understand this is a fix for primary domains with Proxy URL set?

jwt/jwt.go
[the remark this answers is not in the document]
gkats: When we parse the claims, we set the fields of the struct in a similar way that `UnmarshalJSON` operates. 

We could make a pointer out of a non-pointer, but it's hard to detect that.

I believe that we're better off always requesting a pointer. It's something that constructors usually return so it is expected.

There is no guarantee since the return type is `any`, but at least we can provide a guideline.

http/middleware.go
gkats: Yes, but since this goes down to `jwt.Verify`, the context might not make sense there. 

We can have different params for middleware and `jwt.Verify` though. I'll revise this.

gkats: I was wrong, jwt.Verify does accept a context. :smile: 

Updated.

georgepsarakis: Mostly noted this because it's going to be tough to change afterwards.

gkats: And you were 100% right to do so :pray: :smile:

PR #254, opened 2024-02-19
fix: Don't expose jose types

gkats:
Made sure that no go-jose types are exported from our package. 

Had to add more exported types for JWT claims, so that we can re-use go-jose's unmarshaling features.

We now have the following types for JWT claims:
- **RegisteredClaims** These are 'standard' public JWT claims registered in IANA.
- **Claims** These are Clerk-specific private claims, like session ID and active organization ID.
- **SessionClaims** These are the claims that are included in Clerk issued JWTs. They include RegisteredClaims, Claims and any other Custom keys from custom JWT templates. 

Another change introduced here is that the jwt.Decode function will now expose the Key ID header, if the JWT contains it. 

We need access to the kid header when implementing a custom JWT verification flow and jwt.Decode already does most of the work to parse a token. It makes sense for the function to return the Key ID as well.

Here's a sample flow that illustrates why the Key ID is needed.
```go
func someHandler(w http.ResponseWriter, r *http.Request) {
  token := getTokenFromAuthorizationHeader(r)
  decoded, err := jwt.Decode(r.Context(), &jwt.DecodeParams{
    Token: token,
  })
  if err != nil {
    w.WriteHeader(http.StatusUnauthorized)
    return
  }
  jwks, err := jwks.Get(r.Context(), &jwks.GetParams{})
  if err != nil {
    w.WriteHeader(http.StatusUnauthorized)
    return
  }
  var jwk clerk.JSONWebKey
  for _, k := range jwks.Keys {
    if k.KeyID == kid {
      jwk = &k
      break
    }
  }
  if jwk == nil {
    w.WriteHeader(http.StatusUnauthorized)
    return
  }
  claims, err := jwt.Verify(r.Context(), &jwt.VerifyParams{
    Token: token,
     JWK: jwk,
  })
  if err != nil {
    w.WriteHeader(http.StatusUnauthorized)
    return
  }
  w.Write([]byte(claims.SessionID))
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,33 +139,35 @@ import (
     "github.com/clerk/clerk-sdk-go/v2/user"
 )
 
-// Each operation requires a context.Context as the first argument.
-ctx := context.Background()
+func main() {
+    // Each operation requires a context.Context as the first argument.
+    ctx := context.Background()
 
-// Set the API key
-clerk.SetKey("sk_live_XXX")
+    // Set the API key
+    clerk.SetKey("sk_live_XXX")
 
-// Create an organization
-org, err := organization.Create(ctx, &organization.CreateParams{
-    Name: clerk.String("Clerk Inc"),
-})
+    // Create an organization
+    org, err := organization.Create(ctx, &organization.CreateParams{
+        Name: clerk.String("Clerk Inc"),
+    })
 
-// Update the organization
-org, err = organization.Update(ctx, org.ID, &organization.UpdateParams{
-    Slug: clerk.String("clerk"),
-})
+    // Update the organization
+    org, err = organization.Update(ctx, org.ID, &organization.UpdateParams{
+        Slug: clerk.String("clerk"),
+    })
 
-// List organization memberships
-listParams := organizationmembership.ListParams{}
-listParams.Limit = clerk.Int64(10)
-memberships, err := organizationmembership.List(ctx, params)
-if memberships.TotalCount < 0 {
-    return
+    // List organization memberships
+    listParams := organizationmembership.ListParams{}
+    listParams.Limit = clerk.Int64(10)
+    memberships, err := organizationmembership.List(ctx, params)
+    if memberships.TotalCount < 0 {
+        return
+    }
+    membership := memberships[0]
+
+    // Get a user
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -139,33 +139,35 @@ import (
...
-
-// List organization memberships
-listParams := organizationmembership.ListParams{}
-listParams.Limit = clerk.Int64(10)
-memberships, err := organizationmembership.List(ctx, params)
-if memberships.TotalCount < 0 {
-    return
+    // Get a user
+    usr, err := user.Get(ctx, membership.UserID)
 }
```
gkats, line 170: This change is unrelated. I just wrapped the code snippet in a main() function.

jwt.go
```diff
@@ -62,10 +69,70 @@ func (s *SessionClaims) HasRole(role string) bool {
 	return s.ActiveOrganizationRole == role
 }
 
-// Claims holds generic JWT claims.
+// RegisteredClaims holds public claim values (as specified in RFC 7519).
+type RegisteredClaims struct {
```
chanioxaris, line 73: ❓ Instead of retyping/copying the struct, what about create an alias instead?
```golang
type RegisteredClaims jwt.Claims
```

gkats: I'm not sure if it's better to add a type alias or break away completely from go-jose and redefine everything on our own.

Adding a type alias means that our type can be cast to the go-jose type and this way we can use all of go-jose's methods. Is this something we want? Could be a benefit or not.

It also enables us to take advantage of json unmarshaling without overriding any methods. This is clearly a benefit. The less code we have to write, the better.

However, if we redefine everything ourselves, we completely hide go-jose. There's no way you can use the unexported go-jose field and there's no need to ever reach for it.

I'm happy to change this if you think it's better. I don't have a strong opinion. :smile:

georgepsarakis: If you define a type alias I believe someone can still do `jwt.Claims(claims)` thus potentially going against the encapsulation we're trying to achieve here. I think embedding/unexporting the jose struct might be inevitable.

georgepsarakis (approved):

jwt.go
chanioxaris: Thanks for the explanation, makes sense to me!

PR #257, opened 2024-02-21
fix: Concise upgrade instructions

gkats:
Changed the UPGRADING guide for the v1 to v2 migration to be more concise and mention specifically what changed need to be done. 

Left the old content as a "v2 migration guide" document, which explains the reasoning behind the rewrite with more details.

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -32,178 +32,48 @@ API operations in `v1` of the Clerk Go SDK are organized by service. There are
 different services for each API and all services are properties of a single
 `clerk.Client`.
 
-Let's see a concrete example. Assume that we want to create an organization
-and then list all available organizations.
+In `v2` API operations are grouped by API resource. Every API resource is defined in its own package.
 
-Here's how we would do it in `v1`.
-
-```go
-import (
-    "github.com/clerkinc/clerk-sdk-go"
-)
-
-
-func main() {
-    client, err := clerk.NewClient("sk_live_XXX")
-    if err != nil {
-        // handle error
-    }
-
-    // Create an organization
-    org, err := client.Organizations().Create(clerk.CreateOrganizationParams{
-        Name: "Acme Inc",
-    })
-    if err != nil {
-        if errResp, ok := err.(*clerk.ErrorResponse); ok {
-            // Access the API errors
-            errResp.Errors
-        }
-    }
-    // List all organizations, limit results to one.
-    limit := 1
-    orgs, err := client.Organizations().ListAll(clerk.ListAllOrganizationsParams{
-        Limit: &limit,
-    })
-    if err != nil {
-        // handle the error
-    }
-    if orgs.TotalCount > 0 {
-        // Get the first organization in the list
-        org = orgs.Data[0]
-    }
-}
+```diff
+// Create an organization, in v1 and v2. Error handling is omitted.
+- client, err := clerk.NewClient("sk_live_XXX")
+- org, err := client.Organizations().Create(clerk.CreateOrganizationParams{
+-     Name: "Acme Inc",
+- })
++ ctx := context.Background()
++ clerk.SetKey("sk_live_XXX")
++ org, err := organization.Create(ctx, &organization.CreateParams{
++     Name: clerk.String("Acme Inc"),
++ })
 ```
 
-The `v2` version breaks away of the client and services pattern seen above.
-In `v2`, every API has its own package, following a resource-based structure.
+### Support for context.Context
 
-There are two ways to call API operations in `v2`: with or without a client. Let's see both approaches.
-
-### Usage without a client
-
-In most cases, you'll only have to deal with a single API key in your project.
-
-```go
-import (
-    "context"
-
-    "github.com/clerk/clerk-sdk-go/v2"
-    "github.com/clerk/clerk-sdk-go/v2/organization"
-)
-
-func main() {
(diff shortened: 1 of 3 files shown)


UPGRADING.md
```diff
@@ -32,178 +32,48 @@ API operations in `v1` of the Clerk Go SDK are organized by service. There are
 different services for each API and all services are properties of a single
 `clerk.Client`.
 
-Let's see a concrete example. Assume that we want to create an organization
```
gkats, line 35: To the reviewer: It might be easier to click "View file" from the top right menu and see the final outcome in preview mode.

kostaspt (approved):

v2_migration_guide.md
```diff
@@ -0,0 +1,353 @@
+# Migration guide for v2
```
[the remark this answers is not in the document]
chanioxaris: I believe it does! Not every user will have a look at this document as they probably dive in the code directly, but for others it's a helpful hand

chanioxaris (approved):

UPGRADING.md
```diff
@@ -218,111 +88,64 @@ domain.Create(context.Background(), &domain.CreateParams{
...
-
-func handleSession(w http.ResponseWriter, r *http.Request) {
-    sessionClaims, ok := clerk.SessionFromContext(r.Context())
-    if ok {
-        // claims contain session information
-    } else {
-        // there is no active session (non-authenticated user)
-    }
-}
+### The type for failed API responses is now `*clerk.APIErrorResponse`
```
georgepsarakis, line 93:
```suggestion
### Error Response type changed to `clerk.APIErrorResponse`
```

UPGRADING.md
```diff
@@ -218,111 +88,64 @@ domain.Create(context.Background(), &domain.CreateParams{
...
+- }
++ ctx := context.Background()
++ clerk.SetKey("sk_live_XXX")
++ org, err := organization.Create(ctx, &organization.CreateParams{
++     Name: clerk.String("Acme Inc"),
++ })
++ if err != nil {
++     if apiErr, ok := err.(*clerk.APIErrorResponse); ok {
++         // Access the API errors and additional information
++         apiErr.TraceID
```
georgepsarakis, line 118: 💡 You can maybe emulate logging here, so that this section becomes valid Go as well:

```go
fmt.Printf("[%s] request failed: %s (%s)\n", 
  apiErr.TraceID,   
  apiErr.Error(), 
  string(apiErr.Response.RawJSON))
```

UPGRADING.md
```diff
@@ -218,111 +88,64 @@ domain.Create(context.Background(), &domain.CreateParams{
...
-    } else {
-        // there is no active session (non-authenticated user)
-    }
-}
-```
+Please note that as the name implies `WithHeaderAuthorization` and `RequireHeaderAuthorization` support only authentication with a bearer token.
+The token needs to be provided in the "Authorization" request header.
 
-If you're dealing with multiple Clerk API keys, you can pass a `jwks.Client` to the `RequireHeaderAuthorization` middleware.
+**Cookie based authentication is not supported at all by the `v2` version of the library.**
```
georgepsarakis, line 132: 💡 For tips and warnings you might benefit from the GHM blockquote syntax: https://docs.github.com/en/enterprise-cloud@latest/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts

v2_migration_guide.md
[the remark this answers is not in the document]
georgepsarakis: 💡 Just an idea to maybe rename it to `v1_to_v2_migration_guide.md`.

v2_migration_guide.md
```diff
@@ -0,0 +1,353 @@
...
+All types were consolidated and updated to match the currently supported request parameters and responses
+of the [Clerk Backend API](https://clerk.com/docs/reference/backend-api). Many fields have been renamed and a lot
+of new fields have been added. Deprecated struct fields have been dropped.
+
+## Operation parameters
+
+The `v2` version of Clerk SDK Go introduces another important change about types that can be used as API operation
+parameters. Every field for these structs is a pointer.
+
+The `v2` version of the library introduces helper functions to cast basic type values to pointers. These are:
```
georgepsarakis, line 189: ❓ Not sure if `cast` is the right verb here 🤔 I see `generates a pointer` here:

https://go.dev/tour/moretypes/1

georgepsarakis (approved):
👏

UPGRADING.md
gkats: Nice! I think I'll change it altogether to remove code and focus on just the type change.

The problem with Go is that it's too verbose and it's hard to draw the line of how many details to add while preserving readability on the change in focus.

v2_migration_guide.md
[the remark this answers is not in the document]
gkats: Well, I guess you can only migrate from v1, but yes, sure!

UPGRADING.md
gkats: Sweet, TIL and used that!

v2_migration_guide.md
gkats: Updated.

UPGRADING.md
gkats: Updated.

PR #260, opened 2024-02-22
fix: JWKS API uses client

gkats:
Fixed a bug where the jwks.Client.Get method would use the default Backend instead of the one attached to the client.


chanioxaris (approved):

jwks/client.go
```diff
@@ -36,6 +36,6 @@ func (c *Client) Get(ctx context.Context, params *GetParams) (*clerk.JSONWebKeyS
 	req := clerk.NewAPIRequest(http.MethodGet, path)
 	req.SetParams(params)
 	resource := &clerk.JSONWebKeySet{}
-	err := clerk.GetBackend().Call(ctx, req, resource)
```
georgepsarakis, line 39: 💯 Nice catch!

PR #261, opened 2024-02-22
fix: Auth middleware params should not be mutated

gkats:
Applying authentication options should happen inside the handler to avoid the authorization parameters from getting mutated on each handler execution.

This was caught by a test in an upcoming PR. Sadly, it's hard to write a test to catch the bug here.

### DIFF
--- a/http/middleware.go
+++ b/http/middleware.go
@@ -36,20 +36,15 @@ func RequireHeaderAuthorization(opts ...AuthorizationOption) func(http.Handler) 
 // is expected to have the following format:
 // Authorization: Bearer <token>
 func WithHeaderAuthorization(opts ...AuthorizationOption) func(http.Handler) http.Handler {
-	var paramsErr error
-	params := &AuthorizationParams{}
-	for _, opt := range opts {
-		paramsErr = opt(params)
-		if paramsErr != nil {
-			break
-		}
-	}
-
 	return func(next http.Handler) http.Handler {
 		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-			if paramsErr != nil {
-				w.WriteHeader(http.StatusUnauthorized)
-				return
+			params := &AuthorizationParams{}
+			for _, opt := range opts {
+				err := opt(params)
+				if err != nil {
+					w.WriteHeader(http.StatusUnauthorized)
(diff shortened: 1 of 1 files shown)


chanioxaris (approved):

http/middleware.go
```diff
@@ -36,20 +36,15 @@ func RequireHeaderAuthorization(opts ...AuthorizationOption) func(http.Handler)
...
-			break
-		}
-	}
-
 	return func(next http.Handler) http.Handler {
 		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-			if paramsErr != nil {
-				w.WriteHeader(http.StatusUnauthorized)
-				return
+			params := &AuthorizationParams{}
```
georgepsarakis, line 41: 💡 Should we maybe also avoid passing pointers to params fields such as in line 70?

```go
claims, err := jwt.Verify(r.Context(), &params.VerifyParams)
```

gkats: I thought about it as well, but in this case at least I believe it's not "our" fault and the call chain won't modify something surprisingly?

georgepsarakis: Agreed but on the other hand is there any point in passing a pointer? 🤔 Raising this now since we won't be able to change this later, it's a breaking change.

gkats: Do you mean changing all the SDK's methods to accept parameters as non-pointer types?

georgepsarakis: Unless the `nil` check is necessary or convenient, generally I don't see benefit in using pointers but I may be missing something. I was mostly referring to `jwt.Verify` though since it can encounter conditions where parameter objects are propagated through multiple layers. I can't imagine other methods having the same kind of flow, parameters will be probably constructed just near the call!

gkats: I see, thanks for the explanation. I understand the concern, but I'm not sure what's best from a design perspective.

So you're suggesting that all SDK methods accept pointers except jwt.Verify? What about jwt.Decode? Does it fall in the same category?

What's the general heuristic for deciding whether we should accept a pointer or non-pointer?

Is it ok to have all SDK methods accept pointers to structs (what we do now) except for one or two methods?

PR #262, opened 2024-02-23
feat: Same type for API client config

gkats:
All API client constructors get the same type for their configuration. 

Instead of having a ClientConfig type per API package, we use a central clerk.ClientConfig type.


alex-ntousias (approved):
🙏

PR #263, opened 2024-02-23
feat: Clock interface as a time source

gkats:
Verifying a session JWT includes time sensitive validations, like checking token expiration, or that the token can be used now ('nbf' claim).

The HTTP middleware caches JSON web keys for one hour. After that, the cache is considered invalid and a new request to fetch the JSON web key set will be issued.

These are both flows that depend on the current time and time comparisons.

Various system architectures with a distributed nature might not be synchronized. A clock skew is expected.
In such cases it's common to share a "time source" so that all nodes can coordinate on what the perceived current time is.

This commit adds a new clerk.Clock interface which can be passed as a dependency in the http middleware functions and the jwt.Verify function. We also add a dependency to the clockwork project, in order to use a fake clock for our tests.

### DIFF
--- a/clerk.go
+++ b/clerk.go
@@ -448,6 +448,32 @@ type ClientConfig struct {
 	BackendConfig
 }
 
+// Clock is an interface that can be used with the library instead
+// of the [time] package.
+// The interface is useful for testing time sensitive paths or
+// feeding the library with an authoritative source of time, like
+// an external time generator.
+type Clock interface {
+	Now() time.Time
+}
+
+// A default implementation of a Clock, keeping the real time by
+// using the [time] package directly.
+type defaultClock struct{}
+
+// Now returns the current time.
+func (c *defaultClock) Now() time.Time {
+	return time.Now()
+}
+
+// NewClock returns a default clock implementation which calls
+// the [time] package internally.
+// Please note that the return type is an interface because the
+// Clock is not supposed to be used directly.
+func NewClock() Clock {
+	return &defaultClock{}
+}
+
 // Regular expression that matches multiple backslashes in a row.
 var extraBackslashesRE = regexp.MustCompile("([^:])//+")
 
--- a/go.mod
+++ b/go.mod
@@ -9,6 +9,7 @@ require (
 
 require (
 	github.com/davecgh/go-spew v1.1.1 // indirect
+	github.com/jonboulle/clockwork v0.4.0 // indirect
 	github.com/pmezard/go-difflib v1.0.0 // indirect
 	golang.org/x/crypto v0.0.0-20190911031432-227b76d455e7 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
--- a/http/middleware.go
+++ b/http/middleware.go
@@ -46,6 +46,9 @@ func WithHeaderAuthorization(opts ...AuthorizationOption) func(http.Handler) htt
 					return
 				}
 			}
+			if params.Clock == nil {
+				params.Clock = clerk.NewClock()
+			}
 
 			authorization := strings.TrimSpace(r.Header.Get("Authorization"))
 			if authorization == "" {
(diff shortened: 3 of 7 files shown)


clerk.go
```diff
@@ -448,6 +448,32 @@ type ClientConfig struct {
 	BackendConfig
 }
 
+// Clock is an interface that can be used with the library instead
+// of the [time] package.
```
gkats, line 452: To the reviewer: The `Clock` interface is not a drop-in replacement for the [time](https://pkg.go.dev/time) package.

I've only defined one method on `Clock`, `Now()`, because this is the only function of the `time` package that our library currently needs.

I'm not certain whether we should keep the interface small, or require that it implements all of the `time` package's available functions, even if we don't currently need them.

The alternative would be to have this interface for `clerk.Clock`:
```go
type Clock interface {
	After(d time.Duration) <-chan time.Time
	Sleep(d time.Duration)
	Now() time.Time
	Since(t time.Time) time.Duration
	NewTicker(d time.Duration) Ticker
	NewTimer(d time.Duration) Timer
	AfterFunc(d time.Duration, f func()) Timer
}
```

Let me know what you think.

go.mod
```diff
@@ -9,6 +9,7 @@ require (
 
 require (
 	github.com/davecgh/go-spew v1.1.1 // indirect
+	github.com/jonboulle/clockwork v0.4.0 // indirect
```
gkats, line 12: Added this external dependency, but it's used only in tests. Would be nice to implement our own fake clock. At least there are no other indirect dependencies, the clockwork library is small.

http/middleware.go
```diff
@@ -280,6 +295,15 @@ type cacheEntry struct {
 	expiresAt time.Time
 }
 
+// IsValid returns true if a non-expired entry exists in the cache
+// for the provided key, false otherwise.
+func (c *jwkCache) IsValid(key string, t time.Time) bool {
```
georgepsarakis, line 300: 🔧 Minor but perhaps `currentTime` or `now` makes a bit of sense as an argument name?

```suggestion
func (c *jwkCache) IsValid(key string, now time.Time) bool {
```

http/middleware.go
```diff
@@ -84,20 +87,21 @@ func WithHeaderAuthorization(opts ...AuthorizationOption) func(http.Handler) htt
...
 // value.
-func getJWK(ctx context.Context, jwksClient *jwks.Client, kid string) (*clerk.JSONWebKey, error) {
+func getJWK(ctx context.Context, jwksClient *jwks.Client, kid string, clock clerk.Clock) (*clerk.JSONWebKey, error) {
 	if kid == "" {
 		return nil, fmt.Errorf("missing jwt kid header claim")
 	}
 
 	jwk := getCache().Get(kid)
-	if jwk == nil {
+	if jwk == nil || !getCache().IsValid(kid, clock.Now().UTC()) {
```
georgepsarakis, line 96: 💡 Not sure if that's difficult but perhaps the clock should be a field of `jwkCache `? That would solve the issue of checking for expiration inside `Get` if we want to retain this approach.

Actually, maybe ignore that I see that `getCache` is a singleton.

georgepsarakis (approved):

clerk.go
chanioxaris: I feel like the `Now()` method is enough for what we are trying to achieve here, expose a way to DI the clock dependency for testing purposes and not only

clerk.go
```diff
@@ -448,6 +448,32 @@ type ClientConfig struct {
...
+	Now() time.Time
+}
+
+// A default implementation of a Clock, keeping the real time by
+// using the [time] package directly.
+type defaultClock struct{}
+
+// Now returns the current time.
+func (c *defaultClock) Now() time.Time {
+	return time.Now()
```
chanioxaris, line 466: 🔧 I believe we should use UTC by default
```suggestion
	return time.Now().UTC()
```

chanioxaris (approved):

clerk.go
gkats: Thanks. I'll update this in a follow up PR.

gkats: https://github.com/clerk/clerk-sdk-go/pull/267

PR #266, opened 2024-02-26
feat: jwt.Verify fetches the JSON web key

gkats:
In order to reduce friction when manually verifying a session JWT, the jwt.Verify function will fetch the JSON web key if it's not passed in the params.

Users can specify the jwks.Client that will be used to invoke the API GET /v1/jwks method.

Exported the method which fetches the JSON web key, since it can be re-used by the middleware. It can also be used for consumers who wish to fetch the JSON web key once and cache it. Caching is recommended.

```go
// Fetch the JWK once
decoded, _ := jwt.Decode(ctx, &jwt.DecodeParams{Token: token})
jwk, _ := jwt.GetJSONWebKey(ctx, &jwt.GetJSONWebKeyParams{KeyID: decoded.KeyID})

// Use the JWK to verify the token
claims, _ := jwt.Verify(ctx, &jwt.VerifyParams{
  Token: token,
  JWK: jwk,
})
```

### DIFF
--- a/jwt/jwt.go
+++ b/jwt/jwt.go
@@ -9,6 +9,7 @@ import (
 	"time"
 
 	"github.com/clerk/clerk-sdk-go/v2"
+	"github.com/clerk/clerk-sdk-go/v2/jwks"
 	"github.com/go-jose/go-jose/v3/jwt"
 )
 
@@ -23,9 +24,18 @@ type CustomClaimsConstructor func(context.Context) any
 type VerifyParams struct {
 	// Token is the JWT that will be verified. Required.
 	Token string
-	// JWK the custom JSON Web Key that will be used to verify the
-	// Token with. Required.
+	// JWK is the custom JSON Web Key that will be used to verify the
+	// Token with.
+	// If the JWK parameter is provided, the Verify method won't
+	// fetch the JSON Web Key Set and there's no need to provide
+	// the JWKSClient parameter.
 	JWK *clerk.JSONWebKey
+	// JWKSClient is a jwks API client that will be used to fetch the
+	// JSON Web Key Set for verifying the Token with.
+	// If the JWK parameter is provided, the JWKSClient is not needed.
+	// If no JWK or JWKSClient is provided, the Verify method will use
+	// a JWKSClient with the default Backend.
+	JWKSClient *jwks.Client
 	// Clock can be used to keep track of time and will replace usage of
 	// the [time] package. Pass a custom Clock to control the source of
 	// time or facilitate testing chronologically sensitive flows.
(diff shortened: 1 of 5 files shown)


jwt/jwt.go
```diff
@@ -23,9 +24,18 @@ type CustomClaimsConstructor func(context.Context) any
...
+	// If the JWK parameter is provided, the Verify method won't
+	// fetch the JSON Web Key Set and there's no need to provide
+	// the JWKSClient parameter.
 	JWK *clerk.JSONWebKey
+	// JWKSClient is a jwks API client that will be used to fetch the
+	// JSON Web Key Set for verifying the Token with.
+	// If the JWK parameter is provided, the JWKSClient is not needed.
+	// If no JWK or JWKSClient is provided, the Verify method will use
+	// a JWKSClient with the default Backend.
+	JWKSClient *jwks.Client
```
chanioxaris, line 38: ❓ What's the benefit of allowing users to define a jwks client? I mean, can you think of a use case that someone doesn't want to use the already existing backend client?

gkats: There are two ways to use the library.

With a client or without one. The recommended approach which will probably cover most of the cases is without a client.

However, if for some reason you're dealing with multiple API keys, or want more flexible configuration (test mocks, parallel tests), you might prefer to use clients for APIs.

See the README for more details. https://github.com/clerk/clerk-sdk-go/tree/v2?tab=readme-ov-file#usage

alex-ntousias: Why don't we just make the `jwt` client-aware as well?
For example:
```
jwt.Verify(...) // <- create a client, call jwks with that client, and then verify

// or
jwt.NewClient(clientConfig).Verify(....) // <- use the created client to call jwks and then verify
```

gkats: > Why don't we just make the jwt client-aware as well?

@alex-ntousias But the JWKS API already has a client that you can use for JWKS related operations. https://github.com/clerk/clerk-sdk-go/blob/c1f5081d7fb73df94b08c41aa594eb59285382c6/jwks/client.go#L31

What will the `jwt.NewClient` return and how does it relate to the JWKS API?

alex-ntousias: Having a client for `jwt` will help us avoid having to pass a JWKSClient, which is a bit confusing as @chanioxaris mentioned. This should be transparent to the caller.
So, what I'm thinking, is something like this:
```
// in jwt
func (c *Client) Verify(ctx context.Context, params *VerifyParams) (*clerk.SessionClaims, error) {
    if jwk == nil {
        // use jwks to fetch it
        jwksKeySet, err := jwks.NewClient(c.clientConfig).Get(....)
        ....
    }
    ....
}
```

gkats: Thanks for the explanation. I can't say I like this approach, but if it's two against one it probably means it's the right thing to do.

Clients are used for API operations. API operations can also be called without a client.

In this case, we're breaking away from the "rule" for no apparent reason.

Adding a `jwt.Client` that doesn't do anything except passing its config to a `jwks.Client` feels more surprising and confusing than instantiating and passing a `jwks.Client` whose documented use is to fetch the JWK if it's not present.

I understand I might be deep into the problem. Perhaps all this is confusing to fresh eyes.

@alex-ntousias @chanioxaris Since you both reviewed this, please let me know how to proceed so we can complete the work. Thanks!

chanioxaris: The issue here is that `jwt.Verify()` is not part of the client, as it's a utility rather than an API Operation, and for this reason there is no way to inject a custom client, but we will use every time the backend http client by default?

gkats: > it's a utility rather than an API Operation

Yes. We have other "utilities" as well, like jwt.Decode. Decode also doesn't need a "client".

> there is no way to inject a custom client

There is a way to inject a custom jwks.Client, by passing it in the parameters.

```go
jwksClient := jwks.NewClient(config)
jwt.Verify(ctx, &jwt.VerifyParams{
  Token: "",
   JWKSClient: jwksClient,
})
```

> we will use every time the backend http client by default

If you don't pass a jwks.Client in the parameters, we will initialize a jwks.Client with what we consider to be the default configuration.

In most cases, it is expected that users won't need to pass a jwks.Client, unless they:
- deal with multiple API keys in one application
- want parallel testing with no shared Backend

If you need to pass a custom __http__ client (http.Client) for jwt.Verify only and not the rest of your code, you can initialize a jwks.Client and pass it in the parameters.

```go
config := clerk.ClientConfig
config.HTTPClient = &http.Client{}
jwt.Verify(ctx, &jwt.VerifyParams{
  JWKSClient: jwks.NewClient(config),
})
```

chanioxaris: Cool, as far as we are updating our `jwt.Verify()` to dynamically fetch and parse the jwks of the instance in order to validate the token signature, we need a way to inject such a client, mostly for testing purposes. So personally i am good with these changes as i don't expect to be used frequently either way

chanioxaris (approved):

PR #269, opened 2024-02-27
feat: SAML Connections API

chanioxaris:
Added support for the SAML Connections API. The available operations are Create, Update, Read, Delete and List


samlconnection/client.go
```diff
@@ -0,0 +1,119 @@
...
+	Backend clerk.Backend
+}
+
+func NewClient(config *clerk.ClientConfig) *Client {
+	return &Client{
+		Backend: clerk.NewBackend(&config.BackendConfig),
+	}
+}
+
+type AttributeMappingParams struct {
```
chanioxaris, line 26: Note to reviewer: Not sure if we prefer to redefine this struct here in order to be explicit or re-use it from the clerk package

gkats (approved):
:rocket:

PR #264, opened 2024-02-23
chore: add github action for semgrep to run security scans in monitoring mode

zythosec:
Add semgrep in monitoring mode to scan for security issues in the background such as:

- Committed secrets
- First-party code vulnerabilities
- Dependency vulnerabilities

PR #270, opened 2024-02-29
chore: add github action for semgrep to run security scans in monitoring mode

zythosec:
@gkats duplicating #264 for v2


gkats (approved):
:pray:

PR #276, opened 2024-03-22
API base URL and versioning

gkats:
The /v1/ part of API endpoint paths will no longer signal the API version. It can be considered part of the base API URL.

All requests set the Clerk-API-Version header to pin the SDK to a specific API version.


Nikpolik (approved):
🚀

PR #278, opened 2024-03-28
fix: Trim only trailing slashes in URLs

gkats:
Updated the clerk.JoinPath method to trim slashes only at the end of URLs.


clerk.go
```diff
@@ -275,7 +275,6 @@ func (b *defaultBackend) newRequest(ctx context.Context, apiReq *APIRequest) (*h
 	req.Header.Add("Clerk-API-Version", clerkAPIVersion)
 	req.Header.Add("X-Clerk-SDK", fmt.Sprintf("go/%s", sdkVersion))
 	b.CustomRequestHeaders.apply(req)
-	req = req.WithContext(ctx)
```
gkats, line 278: This change is unrelated, but we're already instantiating a request with the context above by calling `http.NewRequestWithContext` so I removed it.

mzhong9723 (approved):

PR #283, opened 2024-04-15
feat: Custom authorization failure handler

### LINKED ISSUE #277: VerifyToken and getJWK without middleware in v2
I am currently working on migrating to v2.

But we cannot use this middleware directly as we want to implement other things inside the Middleware.
https://github.com/clerk/clerk-sdk-go/blob/39301db1dd08e6196e35666ed8f9cb338befb92a/http/middleware.go#L38-L84

However, we want to implement jwt.VerifyToken and getJWK. Since getJWK is not public, we cannot use it as is. We can achieve the same thing by copying the implementation of getJWK, but how does the Clerk team recommend implementing it in such cases?

gkats:
Added an option to the http.WithHeaderAuthorization middleware to modify the default response in case of authentication failure.

Fixes #277.

### DIFF
--- a/http/middleware_test.go
+++ b/http/middleware_test.go
@@ -13,6 +13,29 @@ import (
 )
 
 func TestWithHeaderAuthorization_InvalidAuthorization(t *testing.T) {
+	kid := "kid-" + t.Name()
+	// Mock the Clerk API server. We expect requests to GET /jwks.
+	clerkAPI := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.URL.Path == "/jwks" && r.Method == http.MethodGet {
+			_, err := w.Write([]byte(
+				fmt.Sprintf(
+					`{"keys":[{"use":"sig","kty":"RSA","kid":"%s","alg":"RS256","n":"ypsS9Iq26F71B3lPjT_IMtglDXo8Dko9h5UBmrvkWo6pdH_4zmMjeghozaHY1aQf1dHUBLsov_XvG_t-1yf7tFfO_ImC1JqSQwdSjrXZp3oMNFHwdwAknvtlBg3sBxJ8nM1WaCWaTlb2JhEmczIji15UG6V0M2cAp2VK_brcylQROaJLC2zVa4usGi4AHzAHaRUTv6XB9bGYMvkM-ZniuXgp9dPurisIIWg25DGrTaH-kg8LPaqGwa54eLEnvfAe0ZH_MvA4_bn_u_iDkQ9ZI_CD1vwf0EDnzLgd9ZG1khGsqmXY_4WiLRGsPqZe90HzaBJma9sAxXB4qj_aNnwD5w","e":"AQAB"}]}`,
+					kid,
+				),
+			))
+			require.NoError(t, err)
+			return
+		}
+	}))
+	defer clerkAPI.Close()
+
+	// Mock the clerk backend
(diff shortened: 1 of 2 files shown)


http/middleware_test.go
```diff
@@ -38,6 +56,19 @@ func TestWithHeaderAuthorization_InvalidAuthorization(t *testing.T) {
 	res, err = ts.Client().Do(req)
 	require.NoError(t, err)
 	require.Equal(t, http.StatusOK, res.StatusCode)
```
georgepsarakis, line 58: ❓ Is this correct? Shouldn't this case return 401 too?

http/middleware_test.go
```diff
@@ -134,6 +165,61 @@ func TestWithHeaderAuthorization_Caching(t *testing.T) {
 	require.Equal(t, 2, totalJWKSRequests)
 }
 
+func TestWithHeaderAuthorization_CustomFailureHandler(t *testing.T) {
+	kid := "kid-" + t.Name()
+	// Mock the Clerk API server. We expect requests to GET /jwks.
+	clerkAPI := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.URL.Path == "/jwks" && r.Method == http.MethodGet {
```
georgepsarakis, line 172: ❓ Same question as above regarding the JWKS endpoint path.

http/middleware_test.go
```diff
@@ -13,6 +13,29 @@ import (
 )
 
 func TestWithHeaderAuthorization_InvalidAuthorization(t *testing.T) {
+	kid := "kid-" + t.Name()
+	// Mock the Clerk API server. We expect requests to GET /jwks.
+	clerkAPI := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.URL.Path == "/jwks" && r.Method == http.MethodGet {
```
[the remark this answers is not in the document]
gkats: API versioning is now determined by [a header or query string parameter](https://clerk.com/docs/backend-requests/versioning/overview#choosing-an-api-version).

As a result, adding the version in the path has no effect and might change in the future.

We decided to think of "v1" as part of the API's base URL, since in the future API endpoint paths are likely to omit it.

https://github.com/clerk/clerk-sdk-go/pull/276

http/middleware_test.go
gkats: As discussed above, see https://github.com/clerk/clerk-sdk-go/pull/276 for details.

http/middleware_test.go
gkats: Back when I first implemented the middleware, I decided to keep the functionality that v1 executes when the bearer token cannot be decoded.

https://github.com/clerk/clerk-sdk-go/blob/59c1229871a66eebbc8bba4cc0855a3da1357e6f/clerk/middleware_v2.go#L60

https://github.com/clerk/clerk-sdk-go/blob/195ed8614b6a20fde4cdb95ff51bccce06976ddb/http/middleware.go#L60

That's why I also added the test case, because it was not intuitive to me as well.

georgepsarakis: Got it, thanks for clarifying!

http/middleware_test.go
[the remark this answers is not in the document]
georgepsarakis: Resolved offline, the test Client is not configured with the `v1/` path prefix, so this is the request path it accesses.

georgepsarakis (approved):

PR #284, opened 2024-04-16
feat: Support reply to email name for custom templates

mzhong9723:


gkats (approved):

clerk/templates.go
```diff
@@ -69,13 +70,15 @@ type UpsertTemplateRequest struct {
 	Markup           string  `json:"markup,omitempty"`
 	Body             string  `json:"body"`
 	FromEmailName    *string `json:"from_email_name"`
+	ReplyToEmailName *string `json:"reply_to_email_name"`
```
gkats, line 73: :question: Should we also `omitempty`?

mzhong9723: @gkats looks like I have to resolve this conversation before merging, so I'll go ahead and do so! Already updated to use `omitempty` here

PR #285, opened 2024-04-16
feat: Support reply to email name for custom templates

mzhong9723:


template.go
```diff
@@ -20,6 +20,7 @@ type Template struct {
 	CanRevert          bool         `json:"can_revert"`
 	CanDelete          bool         `json:"can_delete"`
 	FromEmailName      *string      `json:"from_email_name,omitempty"`
+	ReplyToEmailName   *string      `json:"reply_to_email_name,omitempty"`
```
gkats, line 23: :question: Are there templates with no `reply_to_email_name` or do we always pass a default value? If it's the latter, it might make sense to use a `string` type instead?

Ignore this comment if it's the former. :smile:

gkats (approved):

template.go
mzhong9723: It's the former :)

PR #295, opened 2024-05-15
Allowing the base url to be empty

izaaklauer:
## What changed?

Prior to this, if the [clerk backend config url](https://github.com/clerk/clerk-sdk-go/blob/2d117c9e9511d5e460a408ff958af45c179d2b1e/clerk.go#L182) was left as empty, the SDK would prepend an extra `/` to all paths. So a request to `/foo` would become `//foo`.

This change updates our slash-stripping regex to remove leading slashes from the final URL.

### DIFF
--- a/clerk.go
+++ b/clerk.go
@@ -474,8 +474,8 @@ func NewClock() Clock {
 	return &defaultClock{}
 }
 
-// Regular expression that matches multiple backslashes in a row.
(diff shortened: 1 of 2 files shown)


NicolasLopes7 (approved):
Nice catch @izaaklauer! 👏

gkats (approved):
:rocket:

PR #298, opened 2024-05-29
Add True-Client-IP custom header to be passed along to BAPI

NicolasLopes7:
This PR adds the `True-Client-IP` header on the custom headers map in order to pass it along our APIs. We are introducing request info on the webhook and that will enable us to know who has called other than the user-agent.

### DIFF
--- a/clerk.go
+++ b/clerk.go
@@ -160,7 +160,8 @@ type Params interface {
 // CustomRequestHeaders contains predefined values that can be
 // used as custom headers in Backend API requests.
 type CustomRequestHeaders struct {
-	Application string
+	Application  string
+	TrueClientIP string
 }
 
 // Apply the custom headers to the HTTP request.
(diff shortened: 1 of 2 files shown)


gkats:
@NicolasLopes7 Could you please provide more information on what this change is about, and why we need it/what problem it currently solves? Thanks!

clerk.go
```diff
@@ -169,6 +170,7 @@ func (h *CustomRequestHeaders) apply(req *http.Request) {
 		return
 	}
 	req.Header.Add("X-Clerk-Application", h.Application)
+	req.Header.Add("True-Client-IP", h.TrueClientIP)
```
georgepsarakis, line 173: 🔧 This is a semi-standard header name and can potentially cause issues if it's being set on the request directly. Better use the `X-Clerk-` namespace here as well?

NicolasLopes7: Done! That makes sense!

clerk.go
```diff
@@ -160,7 +160,8 @@ type Params interface {
 // CustomRequestHeaders contains predefined values that can be
 // used as custom headers in Backend API requests.
 type CustomRequestHeaders struct {
-	Application string
+	Application  string
+	TrueClientIP string
```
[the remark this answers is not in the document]
NicolasLopes7: That makes sense @gkats, great catch! I'll sync with you tomorrow to understand how can I just pass the headers coming along from a request (if possible). Closing this PR!

PR #305, opened 2024-07-19
ORGS-67: Update user serializer to add CreateOrganizationsLimit field

NicolasLopes7:
> WARNING: This only should be merged after [this change](https://github.com/clerk/clerk_go/pull/7124) is deployed to production!

This PR adds the `CreateOrganizationsLimit` field for the User serializer and user update params structs.

### DIFF
--- a/user/client.go
+++ b/user/client.go
@@ -98,6 +98,7 @@ type UpdateParams struct {
 	BackupCodes                      *[]string        `json:"backup_codes,omitempty"`
 	DeleteSelfEnabled                *bool            `json:"delete_self_enabled,omitempty"`
 	CreateOrganizationEnabled        *bool            `json:"create_organization_enabled,omitempty"`
+	CreateOrganizationsLimit         *int             `json:"create_organizations_limit,omitempty"`
 	// Specified in RFC3339 format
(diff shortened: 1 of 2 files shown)


user/client.go
```diff
@@ -98,6 +98,7 @@ type UpdateParams struct {
 	BackupCodes                      *[]string        `json:"backup_codes,omitempty"`
 	DeleteSelfEnabled                *bool            `json:"delete_self_enabled,omitempty"`
 	CreateOrganizationEnabled        *bool            `json:"create_organization_enabled,omitempty"`
+	CreateOrganizationsLimit         *int             `json:"create_organizations_limit,omitempty"`
```
gkats, line 101: :question: Is support for this field coming soon or already available?

I can't find it in our docs: https://clerk.com/docs/reference/backend-api/tag/Users#operation/UpdateUser

gkats (approved):
Approving, but let's make sure we support the field in our update operation in production before we merge.

user/client.go
NicolasLopes7: Not yet, it's pending on this PR https://github.com/clerk/clerk_go/pull/7124! I'll add to the readme a warning

izaaklauer (approved):

PR #306, opened 2024-07-20
feat(users): add organization_id parameter support

bripkens:
# Why
The list all endpoint for users can accept organization IDs as a filter criterion. This is helpful to retrieve users with access to certain organizations.

# What
Add support for `OrganizationIDs string[]` exactly how it has been done for the JS SDK.

For comparison, this is how it is done within the JS SDK:

 - [parameter definition](https://github.com/clerk/javascript/blob/3c8fda28cfc99d6266397b75ebe1e6161192283c/packages/backend/src/api/endpoints/UserApi.ts#L37C5-L37C19]
 - [how the parameter is passed to the API call](https://github.com/clerk/javascript/blob/3c8fda28cfc99d6266397b75ebe1e6161192283c/packages/backend/src/api/endpoints/UserApi.ts#L116-L130)
 - [check how query parameters are transformed](https://github.com/clerk/javascript/blob/3c8fda28cfc99d6266397b75ebe1e6161192283c/packages/backend/src/api/request.ts#L69-L79)

### DIFF
--- a/user/client.go
+++ b/user/client.go
@@ -190,14 +190,18 @@ func (c *Client) Delete(ctx context.Context, id string) (*clerk.DeletedResource,
 type ListParams struct {
 	clerk.APIParams
 	clerk.ListParams
-	OrderBy           *string  `json:"order_by,omitempty"`
-	Query             *string  `json:"query,omitempty"`
-	EmailAddresses    []string `json:"email_address,omitempty"`
-	ExternalIDs       []string `json:"external_id,omitempty"`
-	PhoneNumbers      []string `json:"phone_number,omitempty"`
-	Web3Wallets       []string `json:"web3_wallet,omitempty"`
-	Usernames         []string `json:"username,omitempty"`
-	UserIDs           []string `json:"user_id,omitempty"`
+	OrderBy        *string  `json:"order_by,omitempty"`
+	Query          *string  `json:"query,omitempty"`
+	EmailAddresses []string `json:"email_address,omitempty"`
+	ExternalIDs    []string `json:"external_id,omitempty"`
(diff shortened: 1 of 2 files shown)


clerk/users.go
```diff
@@ -108,14 +108,18 @@ func (s *UsersService) Create(params CreateUserParams) (*User, error) {
...
-	UserIDs           []string
+	Limit          *int
+	Offset         *int
+	EmailAddresses []string
+	ExternalIDs    []string
+	PhoneNumbers   []string
+	Web3Wallets    []string
+	Usernames      []string
+	UserIDs        []string
+	// OrganizationIDs filters users that have memberships to the given organizations. For each organization ID, the
```
bripkens: Copy of the docs from https://clerk.com/docs/references/backend/user/get-user-list

izaaklauer: 💯 I love seeing the docs here in the SDK. As a developer, if I can get the answer to "what is this field" without having to leave the IDE, i'm much more efficient.

It breaks our existing pattern, so others can weigh in on whether we should start doing it now in this PR, wait until we have some programmatic solution to add comments in bulk and keep them up-to-date, or object to the premise 😄 . But I'd be OK with including it here.

izaaklauer (approved):
🚀  Thank you! Very straightforward - your contribution is appreciated.

NicolasLopes7:
Just a heads up that we should be applying new changes on the `v2`, So please point `v2` as the base branch. Thanks for your contribution!

bripkens:
@NicolasLopes7 Change is now rebased on the v2 branch and PR target adjusted.

PR #310, opened 2024-08-07
readme fix: ResponseReader is interface, doesn't need ptr

bgentry:
Following this code example as-is results in an error because `reader` is a pointer to an interface instead of just an interface. This fixes it.


gkats (approved):
Thank you for your contribution @bgentry!!

PR #316, opened 2024-08-30
feat: Support limit & offset to invitations List

nikospapcom:
This PR adds support for limit & offset params in the invitations List


Nikpolik (approved):
🚀

PR #322, opened 2024-09-05
feat: create TOTP

NicolasLopes7:
This PR implements a new method under the user's client to create TOTP

### DIFF
--- a/totp.go
+++ b/totp.go
@@ -1,6 +1,6 @@
 package clerk
 
-// TOTP describes a TOTP (Time-based One-Time Password) for a user.
+// Describes a Time-based One-time Password (TOTP) for a user.
 type TOTP struct {
 	APIResource
 	Object      string   `json:"object"`
--- a/totp.go
+++ b/totp.go
@@ -1,0 +1,14 @@
+package clerk
+
+// TOTP describes a TOTP (Time-based One-Time Password) for a user.
+type TOTP struct {
+	APIResource
+	Object      string   `json:"object"`
+	ID          string   `json:"id"`
+	Secret      *string  `json:"secret"`
+	URI         *string  `json:"uri" `
+	Verified    bool     `json:"verified"`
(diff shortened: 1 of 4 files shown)


totp.go
```diff
@@ -0,0 +1,14 @@
+package clerk
+
+// TOTP describes a TOTP (Time-based One-Time Password) for a user.
```
LauraBeatris: Super nit, feel free to ignore. Not sure if starting with the entity is the pattern but it reads a bit repetitive. 

```suggestion
// Describes a Time-based One-time Password (TOTP) for a user.
```

user/api.go
```diff
@@ -23,12 +23,12 @@ func Update(ctx context.Context, id string, params *UpdateParams) (*clerk.User,
...
 }
 
-// UpdateProfileImage sets or replaces the users's profile image.
+// UpdateProfileImage sets or replaces the user's profile image.
 func UpdateProfileImage(ctx context.Context, id string, params *UpdateProfileImageParams) (*clerk.User, error) {
 	return getClient().UpdateProfileImage(ctx, id, params)
 }
 
-// DeleteProfileImage removes the users's profile image.
+// DeleteProfileImage sets or replaces the user's profile image.
```
LauraBeatris: I might be missing some context here on that feature, but why a `Delete` method would `set or replace`?

LauraBeatris (approved):
Looks good! I just left a question, don't want to block the release here.

mzhong9723 (approved):

user/api.go
NicolasLopes7: Oh, that was from `go gen ./...`, It generates this api file based on the methods from the other clients, I'll fix those and re-run it! Thanks for spotting that

NicolasLopes7: resolved via [5548cc3](https://github.com/clerk/clerk-sdk-go/pull/322/commits/5548cc3627c559f138581078390db4a96d4a319e)

PR #324, opened 2024-09-10
fix(invitations): Support multiple statuses

kostaspt:
In this PR, we update the invitations to allow filtering by multiple statuses instead of just one, since we already support this in Backend API.

### DIFF
--- a/invitation/client.go
+++ b/invitation/client.go
@@ -28,9 +28,9 @@ func NewClient(config *clerk.ClientConfig) *Client {
 type ListParams struct {
 	clerk.APIParams
 	clerk.ListParams
-	OrderBy *string `json:"order_by,omitempty"`
-	Query   *string `json:"query,omitempty"`
-	Status  *string `json:"status,omitempty"`
+	OrderBy  *string  `json:"order_by,omitempty"`
+	Query    *string  `json:"query,omitempty"`
+	Statuses []string `json:"status,omitempty"`
 }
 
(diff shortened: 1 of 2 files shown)


mzhong9723 (approved):

thiskevinwang (approved):

gkats (commented):
This is a breaking change. What version are we planning to release this in?

kostaspt:
> This is a breaking change. What version are we planning to release this in?

It’s for `v2` only. AFAIK, we haven’t officially released `v2` yet. Would this still be considered a breaking change?

mzhong9723:
> This is a breaking change. What version are we planning to release this in?

It would be a different version of the SDK right? The API itself isn't changing.

@kostaspt Maybe we can support the previous parameter as well, and give it priority if it is set.

kostaspt:
> This is a breaking change. What version are we planning to release this in?

We added this in https://github.com/clerk/clerk-sdk-go/commit/84b590a821cb5c2fd40c2e3b5ecc8dc15606aeba, and the last release was commit https://github.com/clerk/clerk-sdk-go/commit/240f3a55b2f13f0a293b9019b91e0b9e7848f0d4 [2.0.8] (one before). So we haven’t released this change yet.

Another safe approach is to follow what @mzhong9723 suggested and deprecate it.

gkats:
@kostaspt If there's no tagged version for v2 with the `Status` field then we're ok to merge. :+1:

PR #328, opened 2024-09-19
feat: Add `include_members_count` for `/organizations/:id`

LauraBeatris:


LauraBeatris:
This has become quite stale, I still have to sync with team on which options are better and then open another PR:
- Modify the current method, cut a new major release
- Add a new method, no breaking change

PR #330, opened 2024-09-26
feat(saml_connection): add  `organization_id` field

NicolasLopes7:
`nil`/`unset` -> no action
Empty value -> Removes the `organization_id`
Valid ID -> Add `organization_id`

### DIFF
--- a/samlconnection/client.go
+++ b/samlconnection/client.go
@@ -72,9 +72,9 @@ type UpdateParams struct {
 	Domain      *string `json:"domain,omitempty"`
 	IdpEntityID *string `json:"idp_entity_id,omitempty"`
 	// OrganizationID is a nullable optional field.
-	// - If nil or unset, the backend will not take any action.
-	// - If an empty value (""), the backend will unset the organization_id.
-	// - If a valid ID is provided, the backend will update the organization_id.
+	// - If nil or unset, no action will be taken.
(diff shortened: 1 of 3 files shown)


samlconnection/client.go
```diff
@@ -67,9 +68,14 @@ func (c *Client) Get(ctx context.Context, id string) (*clerk.SAMLConnection, err
...
-	Name                             *string                 `json:"name,omitempty"`
-	Domain                           *string                 `json:"domain,omitempty"`
-	IdpEntityID                      *string                 `json:"idp_entity_id,omitempty"`
+	Name        *string `json:"name,omitempty"`
+	Domain      *string `json:"domain,omitempty"`
+	IdpEntityID *string `json:"idp_entity_id,omitempty"`
+	// OrganizationID is a nullable optional field.
+	// - If nil or unset, the backend will not take any action.
+	// - If an empty value (""), the backend will unset the organization_id.
+	// - If a valid ID is provided, the backend will update the organization_id.
```
mzhong9723: 🙃 
```suggestion
	// - If nil or unset, no action will be taken.
	// - If an empty value (""), the organization_id will be unset.
	// - If a valid ID is provided, the organization_id will be updated.
```

mzhong9723 (approved):

PR #331, opened 2024-09-26
feat(user): add methods to delete totp and backup_code

NicolasLopes7:


user/client.go
```diff
@@ -464,3 +464,27 @@ func (c *Client) CreateTOTP(ctx context.Context, userID string) (*clerk.TOTP, er
...
 }
+
+// DeleteTOTP deletes all the TOTPs from a given user.
+func (c *Client) DeleteTOTP(ctx context.Context, userID string) (*MultifactorAuthentication, error) {
+	path, err := clerk.JoinPath(path, userID, "/totp")
+	if err != nil {
+		return nil, err
+	}
+	req := clerk.NewAPIRequest(http.MethodDelete, path)
+	resource := &MultifactorAuthentication{}
```
mzhong9723, line 475: ❓ It's intentional that this isn't the usual `DeletedResource` type right?

mzhong9723 (approved):

user/client.go
NicolasLopes7: Yep, usually it is, but for these MFAs endpoints we're deleting all the records from a userID, so we are returning just the `userID` (I found a bit odd too, but that's how it was) https://clerk.com/docs/reference/backend-api/tag/Users#operation/DisableMFA

PR #333, opened 2024-10-01
DASH-416: Include the latest activity on session response

NicolasLopes7:
- **feat: add latest activity field on session**


mzhong9723 (approved):
Do we need to update or add any tests?

gkats (approved):

PR #336, opened 2024-10-08
feat(organizationinvitation): implement method to list organization invitations from a given instance

NicolasLopes7:


organizationinvitation/client.go
```diff
@@ -115,3 +115,42 @@ func (c *Client) Revoke(ctx context.Context, params *RevokeParams) (*clerk.Organ
...
+
+	if p.OrderBy != nil {
+		q.Add("order_by", *p.OrderBy)
+	}
+
+	return q
+}
+
+// ListAllFromInstance lists all the organization invitations from the current instance
+func (c *Client) ListFromInstance(ctx context.Context, params *ListFromInstanceParams) (*clerk.OrganizationInvitationList, error) {
```
gkats, line 146: :question: Have we considered reusing the existing `organizationinvitation.List` method?

We could return all organization invitations for an instance if `ListParams.OrganizationID == ""`. Granted, the API would be better if the `OrganizationID` was a pointer to a string, but perhaps a single `List` method reduces the API surface?

What do you think?

NicolasLopes7: I think the issue here is that we have some additional paramaters that are not contained on the one with organizationID. I can implement it to match both though!

gkats: > I can implement it to match both though!

Not sure which is better. Feel free to follow your instinct or ask the rest of the team for feedback. :smile:

NicolasLopes7: I will!

NicolasLopes7: For now I think it's best to have both

izaaklauer (approved):

PR #338, opened 2024-10-13
feat: add role_name field for organization invitation and membership

NicolasLopes7:
This pull request updates the `organization_invitation` and `organization_membership` models to include the `role_name` field


gkats (approved):
Could you provide an example of a response with both "role" and "role_name" fields set?

PR #340, opened 2024-10-22
add sign-in token api

PhilippBuschhaus:
Add the sign-in token api: 
https://clerk.com/docs/reference/backend-api/tag/Sign-in-Tokens


sign_in_token.go
```diff
@@ -0,0 +1,13 @@
+package clerk
+
+type SignInToken struct {
+	APIResource
+	Object    string `json:"object"`
+	ID        string `json:"id"`
+	Status    string `json:"status"`
+	UserID    string `json:"user_id"`
+	Token     string `json:"token,omitempty"`
+	URL       string `json:"url,omitempty"`
```
gkats: I think the URL can also be `null`.

```suggestion
	URL       *string `json:"url,omitempty"`
```

signintokens/client.go
```diff
@@ -0,0 +1,51 @@
+// Package session provides the Sign In Token API.
+package signintokens
```
gkats: :wrench: Let's change the package name (and the folder) to `signintoken`. Most of the package names in the library use singular form for the resources.

signintokens/client.go
```diff
@@ -0,0 +1,51 @@
+// Package session provides the Sign In Token API.
```
gkats:
```suggestion
// Package signintoken provides the Sign In Token API.
```

gkats (changes requested):
Thank you so much for the contribution @PhilippBuschhaus!

Overall the code looks great. I've left a few comments, could you please take another look and then we can :shipit:.

PhilippBuschhaus:
> Thank you so much for the contribution @PhilippBuschhaus!
> 
> Overall the code looks great. I've left a few comments, could you please take another look and then we can :shipit:.

Thanks for the feedback :). I've addressed the issues but let me know if anything else needs changing.

gkats (approved):

MohamedArif20091999:
@PhilippBuschhaus I could see this code is merged, But With latest 	"github.com/clerk/clerk-sdk-go/v2" I'm not able to see this module, Is there something I'm missing here?

PR #348, opened 2024-10-31
feat: Updates User List to support created_at & last_active_at date filters

BrandonRomano:
Add `CreatedAtBefore`, `CreatedAtAfter`, `LastActiveAtBefore` and `LastActiveAtAfter` to `user.ListParams`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## 2.1.1
+## Next release
 
 - Add `EmailAddressQuery`, `PhoneNumberQuery` and `UsernameQuery` to `user.ListParams`.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 2.1.1
+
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 2.1.1
```
gkats: :100: Thank you for adding a changelog entry!

:question: Should we avoid adding a header for the version until we actually release it?

The next version bump could also be minor (not patch). Is it better if we leave the entries at the top without a header? I guess it doesn't matter much, but it would be good to follow a convention among us.

Feel free to leave as is.

gkats (approved):
:rocket:

CHANGELOG.md
BrandonRomano: > ❓ Should we avoid adding a header for the version until we actually release it?

You know, this is a good idea.  I thought of this exact problem when I was typing up this header.

Perhaps we can use the header `## Next release`, just so it's contextualized to anyone who peeks the changelog?

PR #351, opened 2024-11-12
feat: Add bulk invitation creation support

kostaspt:
Ref: https://clerk.com/docs/reference/backend-api/tag/Invitations#operation/CreateBulkInvitations

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Next release
 
+- Add `BulkCreate` method for bulk invitation creation.
+
 ## 2.1.1
 
 - Add `EmailAddressQuery`, `PhoneNumberQuery` and `UsernameQuery` to `user.ListParams`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Next release
 
-- Add `BulkCreate` method for bulk invitation creation.
+- Add support for bulk invitation creation with the `invitation.BulkCreate` method.
 
 ## 2.1.1
 
--- a/invitation/client.go
+++ b/invitation/client.go
@@ -96,12 +96,16 @@ func (b *bulkCreateResponse) UnmarshalJSON(data []byte) error {
 
 // BulkCreate creates multiple invitations.
 func (c *Client) BulkCreate(ctx context.Context, params *BulkCreateParams) (*clerk.Invitations, error) {
-	req := clerk.NewAPIRequest(http.MethodPost, fmt.Sprintf("%s/bulk", path))
+	path, err := clerk.JoinPath(path, "bulk")
+	if err != nil {
+		return nil, err
+	}
+
+	req := clerk.NewAPIRequest(http.MethodPost, path)
 	req.SetParams(params)
 
 	res := &bulkCreateResponse{}
-	err := c.Backend.Call(ctx, req, res)
-	if err != nil {
+	if err := c.Backend.Call(ctx, req, res); err != nil {
 		return nil, err
 	}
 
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Next release
 
+- Add `BulkCreate` method for bulk invitation creation.
```
gkats: :upside_down_face: 

```suggestion
- Add support for bulk invitation creation with the `invitation.BulkCreate` method.
```

invitation/client.go
```diff
@@ -76,6 +76,42 @@ func (c *Client) Create(ctx context.Context, params *CreateParams) (*clerk.Invit
...
+	Invitations []*clerk.Invitation
+}
+
+func (b *BulkCreateResponse) UnmarshalJSON(data []byte) error {
+	return json.Unmarshal(data, &b.Invitations)
+}
+
+// BulkCreate creates multiple invitations.
+func (c *Client) BulkCreate(ctx context.Context, params *BulkCreateParams) (*clerk.InvitationList, error) {
+	req := clerk.NewAPIRequest(http.MethodPost, fmt.Sprintf("%s/bulk", path))
```
gkats: :wrench: Let's use clerk.JoinPath here.

invitation/client.go
```diff
@@ -76,6 +76,42 @@ func (c *Client) Create(ctx context.Context, params *CreateParams) (*clerk.Invit
...
+	req.SetParams(params)
+
+	res := &BulkCreateResponse{}
+	err := c.Backend.Call(ctx, req, res)
+	if err != nil {
+		return nil, err
+	}
+
+	return &clerk.InvitationList{
```
gkats: :question: Could you explain why this and the [UnmarshalJSON](https://github.com/clerk/clerk-sdk-go/pull/351/files#diff-f8054c1dc0c25512da995ae22e82a397a28d629fba1776eafa3b5b2f1eb60b4fR93) method is needed?

Why do we want to return different results than the API?

invitation/client.go
```diff
@@ -76,6 +76,42 @@ func (c *Client) Create(ctx context.Context, params *CreateParams) (*clerk.Invit
 	return invitation, err
 }
 
+type BulkCreateParams struct {
+	clerk.APIParams
+	Invitations []*CreateParams
+}
+
+func (b BulkCreateParams) MarshalJSON() ([]byte, error) {
```
gkats, line 84: :question: Why do we need this custom marshaling method?

kostaspt: We need the custom marshaling because the API's request for bulk is an array of regular requests, so it’s not bound to a key like `{"invitations": [...]}`. To be accurate in SDK, the correct type would be `type BulkCreateParams []*CreateParams`. However this wouldn’t be compatible with `req.SetParams`.

Do you have a better alternative to suggest?

invitation/client.go
kostaspt: Similar to [this thread](https://github.com/clerk/clerk-sdk-go/pull/351#discussion_r1838247796), this can actually be improved if we introduce another type. However, because this cannot be paginated because it's a response, it’s not under a key, it’s just an array. So we would lose the `APIResource` in the process.

invitation/client.go
gkats: Thanks for explaining! It is indeed a problem with the current implementation.

> Do you have a better alternative to suggest?

Not really. :disappointed:  As discussed offline, let's settle on the exported types for now and solidify the API.

kostaspt: I’ve refactored the types as we discussed a bit [here](https://github.com/clerk/clerk-sdk-go/pull/351/commits/0d2d9a136df95219204bcf47d7942ce0bfda3a99). Please take another look when you get a chance.

gkats (approved):
Thanks for spending more time on this one @kostaspt.

Let's not forget to use clerk.JoinPath. Other than that, looks good to me!

BrandonRomano (approved):

PR #353, opened 2024-11-13
feat: Support OAuth Applications 

AlexNti:
In this commit we are adding the CRUD methods and the rotate client secret for  for OAuth Applications, the documentation for those API can be found at https://clerk.com/docs/reference/backend-api/tag/OAuth-Applications


oauthapplication/client.go
```diff
@@ -0,0 +1,116 @@
...
+	Backend clerk.Backend
+}
+
+func NewClient(config *clerk.ClientConfig) *Client {
+	return &Client{
+		Backend: clerk.NewBackend(&config.BackendConfig),
+	}
+}
+
+// fetches a single OAuth application by its ID.
```
gkats:
```suggestion
// Get fetches a single OAuth application by its ID.
```

oauthapplication/client.go
```diff
@@ -0,0 +1,116 @@
...
+	err = c.Backend.Call(ctx, req, resource)
+	return resource, err
+}
+
+type ListParams struct {
+	clerk.APIParams
+	clerk.ListParams
+}
+
+func (params *ListParams) ToQuery() url.Values {
```
gkats, line 43: :question: Is this method needed? Since we're not adding functionality over clerk.ListParams.ToQuery it feels redundant.

oauthapplication/client.go
```diff
@@ -0,0 +1,116 @@
...
+// Retrieves all OAuth applications.
+func (c *Client) List(ctx context.Context, params *ListParams) (*clerk.OAuthApplicationList, error) {
+	req := clerk.NewAPIRequest(http.MethodGet, path)
+	req.SetParams(params)
+	list := &clerk.OAuthApplicationList{}
+	err := c.Backend.Call(ctx, req, list)
+	return list, err
+}
+
+type OAuthApplicationCreateParams struct {
```
gkats: :wrench: 
```suggestion
type CreateParams struct {
```

Also, let's make all fields pointers to types (`*string` and `*bool`).

oauthapplication/client.go
```diff
@@ -0,0 +1,116 @@
...
+	clerk.APIParams
+	Name        string `json:"name"`
+	CallbackURL string `json:"callback_url"`
+	Scopes      string `json:"scopes"`
+	Public      bool   `json:"public"`
+}
+
+// Creates a new OAuth application with the given parameters.
+func (c *Client) Create(ctx context.Context, params *OAuthApplicationCreateParams) (*clerk.OAuthApplicationWithClientSecret, error) {
```
gkats: :question: Do we need the different type here? Why can't `Create` return a `*clerk.OAuthApplication`?

```suggestion
func (c *Client) Create(ctx context.Context, params *CreateParams) (*clerk.OAuthApplication, error) {
```

oauthapplication/client.go
```diff
@@ -0,0 +1,116 @@
...
+
+type OAuthApplicationCreateParams struct {
+	clerk.APIParams
+	Name        string `json:"name"`
+	CallbackURL string `json:"callback_url"`
+	Scopes      string `json:"scopes"`
+	Public      bool   `json:"public"`
+}
+
+// Creates a new OAuth application with the given parameters.
```
gkats:
```suggestion
// Create creates a new OAuth application with the given parameters.
```

oauthapplication/client.go
```diff
@@ -0,0 +1,116 @@
...
+func (c *Client) Create(ctx context.Context, params *OAuthApplicationCreateParams) (*clerk.OAuthApplicationWithClientSecret, error) {
+	req := clerk.NewAPIRequest(http.MethodPost, path)
+	req.SetParams(params)
+	authApplication := &clerk.OAuthApplicationWithClientSecret{}
+	err := c.Backend.Call(ctx, req, authApplication)
+	return authApplication, err
+}
+
+type OAuthApplicationUpdateParams struct {
```
gkats:
```suggestion
type UpdateParams struct {
```

oauthapplication/client.go
```diff
@@ -0,0 +1,116 @@
...
+	clerk.APIParams
+	Name        *string `json:"name,omitempty"`
+	CallbackURL *string `json:"callback_url,omitempty"`
+	Scopes      *string `json:"scopes,omitempty"`
+}
+
+// Updates an existing OAuth application.
```
gkats:
```suggestion
// Update updates an existing OAuth application.
```

oauthapplication/client.go
```diff
@@ -0,0 +1,116 @@
...
+		return nil, err
+	}
+	req := clerk.NewAPIRequest(http.MethodDelete, path)
+	authApplication := &clerk.DeletedResource{}
+	err = c.Backend.Call(ctx, req, authApplication)
+	return authApplication, err
+}
+
+// Rotates the OAuth application's client secret
+func (c *Client) RotateClientSecret(ctx context.Context, id string) (*clerk.OAuthApplicationWithClientSecret, error) {
```
gkats: Let's use one type for `OAuthApplication`s.
```suggestion
func (c *Client) RotateClientSecret(ctx context.Context, id string) (*clerk.OAuthApplication, error) {
```

oauth_application.go
```diff
@@ -0,0 +1,30 @@
...
+	Scopes        string `json:"scopes"`
+	CallbackURL   string `json:"callback_url"`
+	AuthorizeURL  string `json:"authorize_url"`
+	TokenFetchURL string `json:"token_fetch_url"`
+	UserInfoURL   string `json:"user_info_url"`
+	CreatedAt     int64  `json:"created_at"`
+	UpdatedAt     int64  `json:"updated_at"`
+}
+
+type OAuthApplicationWithClientSecret struct {
```
gkats: :question: Do we need an additional type for just one field? How about we add a `*string` field to the `OAuthApplication` type instead?

```go
type OAuthApplication struct {
  ClientSecret *string `json:"client_secret,omitempty"`
}
```

gkats (commented):
Nice one, @AlexNti!

I've left some comments, let me know what you think.

oauth_application.go
AlexNti: You are right this a better approach, I will update the code!

oauthapplication/client.go
AlexNti: Without I am getting the following error, 

`cannot use params (variable of type *ListParams) as clerk.Params value in argument to req.SetParams: *ListParams does not implement clerk.Params (ambiguous selector *ListParams.ToQuery`

Is there any other way to bypass that error?

gkats: I see, the problem is that both ListParams and APIParams implement the ToQuery method. This reveals a bigger problem with our types.

I'd suggest leaving as is for now. :smile:

gkats (approved):
Before you merge this one, could you please also update the [CHANGELOG](https://github.com/clerk/clerk-sdk-go/blob/fe4b64f7cb6df287b94c226fae63faf271483895/CHANGELOG.md)?

You can add a "Next release" with your changes for now. See [this PR](https://github.com/clerk/clerk-sdk-go/pull/350/files) for an example.

PR #362, opened 2024-12-04
Adds support for fetching orgs with `include_members_count`

BrandonRomano:
Adds support for [include_members_count](https://clerk.com/docs/reference/backend-api/tag/Organizations#operation/GetOrganization) when fetching an organization.

I had to introduce a separate function`GetWithParams` as to not introduce a breaking change. This is a good candidate to flatten to a single method for our next major.


NicolasLopes7 (approved):
Looks good to me! And for sure a good candidate for the next major

BrandonRomano:
🙇 Thanks for the quick review ⚡

PR #369, opened 2025-01-17
Fix typo's in the readme

mehiX:
Pass an instance to a roundTripper, not the type


gkats (approved):
Thanks for catching that!

PR #371, opened 2025-01-21
Add missing user.CreateParams properties

joschi:
The `user.CreateParams` struct was missing three (optional) properties which are supported by the [`CreateUser`](https://clerk.com/docs/reference/backend-api/tag/Users#operation/CreateUser) operation:

- `DeleteSelfEnabled`: https://clerk.com/docs/reference/backend-api/tag/Users#operation/CreateUser!path=delete_self_enabled&t=request
- `CreateOrganizationEnabled`: https://clerk.com/docs/reference/backend-api/tag/Users#operation/CreateUser!path=create_organization_enabled&t=request
- `CreateOrganizationsLimit`: https://clerk.com/docs/reference/backend-api/tag/Users#operation/CreateUser!path=create_organizations_limit&t=request

The `user.UpdateParams` struct already contains these properties.

### DIFF
--- a/user/client.go
+++ b/user/client.go
@@ -32,23 +32,26 @@ func NewClient(config *clerk.ClientConfig) *Client {
 
 type CreateParams struct {
 	clerk.APIParams
-	EmailAddresses          *[]string        `json:"email_address,omitempty"`
-	PhoneNumbers            *[]string        `json:"phone_number,omitempty"`
-	Web3Wallets             *[]string        `json:"web3_wallet,omitempty"`
-	Username                *string          `json:"username,omitempty"`
-	Password                *string          `json:"password,omitempty"`
-	FirstName               *string          `json:"first_name,omitempty"`
-	LastName                *string          `json:"last_name,omitempty"`
-	ExternalID              *string          `json:"external_id,omitempty"`
-	UnsafeMetadata          *json.RawMessage `json:"unsafe_metadata,omitempty"`
-	PublicMetadata          *json.RawMessage `json:"public_metadata,omitempty"`
-	PrivateMetadata         *json.RawMessage `json:"private_metadata,omitempty"`
-	PasswordDigest          *string          `json:"password_digest,omitempty"`
-	PasswordHasher          *string          `json:"password_hasher,omitempty"`
-	SkipPasswordRequirement *bool            `json:"skip_password_requirement,omitempty"`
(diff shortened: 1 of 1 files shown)


user/client.go
```diff
@@ -32,23 +32,26 @@ func NewClient(config *clerk.ClientConfig) *Client {
...
+	PrivateMetadata           *json.RawMessage `json:"private_metadata,omitempty"`
+	PasswordDigest            *string          `json:"password_digest,omitempty"`
+	PasswordHasher            *string          `json:"password_hasher,omitempty"`
+	SkipPasswordRequirement   *bool            `json:"skip_password_requirement,omitempty"`
+	SkipPasswordChecks        *bool            `json:"skip_password_checks,omitempty"`
+	TOTPSecret                *string          `json:"totp_secret,omitempty"`
+	BackupCodes               *[]string        `json:"backup_codes,omitempty"`
+	DeleteSelfEnabled         *bool            `json:"delete_self_enabled,omitempty"`
+	CreateOrganizationEnabled *bool            `json:"create_organization_enabled,omitempty"`
+	CreateOrganizationsLimit  *int             `json:"create_organizations_limit,omitempty"`
```
gkats, line 54: Could we change the type here so it's explicitly a pointer to a 64-bit integer? 

Although it's not documented anywhere, we tend to favor typing all integers as `int64`.

```suggestion
	CreateOrganizationsLimit  *int64             `json:"create_organizations_limit,omitempty"`
```

gkats (commented):
Nice addition @joschi!

Left a comment about a struct field type. Could you please take a look? Thanks!

user/client.go
joschi: @gkats I've applied the suggestion but now the `CreateParams` struct is inconsistent with `UpdateParams`.

Should the `UpdateParams` struct be changed, too? Although this would introduce a breaking change in the SDK.

gkats: Oh, I hadn't noticed that we already include these fields in `UpdateParams`, although you did mention it in the PR description. 

You are right, let's favor consistency here and keep the field as `*int` as you had initially written. Is it ok for you to change it back?

Sorry for the confusion, I should have been more careful.

joschi: I've reverted the change of the data type. 😃

gkats: Thank you!

gkats (approved):

PR #373, opened 2025-01-22
fix!: Align OAuth Application specs with API

chanioxaris:
In this commit we are making changes to the OAuth Application resource API. We are aligning the types of the Create operation, to optional, as the only required right now is the 'name' property Also we are exposing a new 'token_introspection_url' property as part of the response. Those are considered breaking changes, that's why they only affect the new major version (v3)

### DIFF
--- a/oauth_application.go
+++ b/oauth_application.go
@@ -2,21 +2,22 @@ package clerk
 
 type OAuthApplication struct {
 	APIResource
-	Object        string  `json:"object"`
-	ID            string  `json:"id"`
-	InstanceID    string  `json:"instance_id"`
-	Name          string  `json:"name"`
-	ClientID      string  `json:"client_id"`
-	Public        bool    `json:"public"`
-	Scopes        string  `json:"scopes"`
-	CallbackURL   string  `json:"callback_url"`
-	DiscoveryURL  string  `json:"discovery_url"`
-	AuthorizeURL  string  `json:"authorize_url"`
-	TokenFetchURL string  `json:"token_fetch_url"`
-	UserInfoURL   string  `json:"user_info_url"`
-	CreatedAt     int64   `json:"created_at"`
-	UpdatedAt     int64   `json:"updated_at"`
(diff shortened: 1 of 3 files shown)


oauth_application.go
```diff
@@ -2,21 +2,22 @@ package clerk
...
+	Scopes                string  `json:"scopes"`
+	CallbackURL           string  `json:"callback_url"`
+	DiscoveryURL          string  `json:"discovery_url"`
+	AuthorizeURL          string  `json:"authorize_url"`
+	TokenFetchURL         string  `json:"token_fetch_url"`
+	UserInfoURL           string  `json:"user_info_url"`
+	TokenIntrospectionURL string  `json:"token_introspection_url"`
+	CreatedAt             int64   `json:"created_at"`
+	UpdatedAt             int64   `json:"updated_at"`
+	ClientSecret          *string `json:"client_secret,omitempty"`
```
kostaspt: 🙃 (nit) Maybe move this closer to `ClientID`?

oauthapplication/client_test.go
```diff
@@ -36,9 +36,9 @@ func TestOAuthApplicationClientCreate(t *testing.T) {
 
 	params := &CreateParams{
 		Name:        name,
-		CallbackURL: callbackURL,
-		Scopes:      scopes,
-		Public:      public,
+		CallbackURL: clerk.String(callbackURL),
+		Scopes:      clerk.String(scopes),
```
kostaspt, line 40: 🙃 What if we forced the scopes to be a `[]string`, since it's a breaking change, and then joined them when creating the params for the Backend API?

kostaspt (approved):

oauthapplication/client_test.go
chanioxaris: Yeah definitely agree here! It's on my list for the next major API version

oauth_application.go
chanioxaris: Good call out!

PR #374, opened 2025-01-27
chore: Make CreateTOTP noop and mark as deprecated

octoper:
The endpoint used for `CreateTOTP` will be removed so this PR makes `CreateTOTP` a noop to avoid code level breaking changes and also marks it as deprecated as it will be removed completely in `v3`

Resolves USER-1109

### DIFF
--- a/organization/api.go
+++ b/organization/api.go
@@ -19,6 +19,12 @@ func Get(ctx context.Context, idOrSlug string) (*clerk.Organization, error) {
 	return getClient().Get(ctx, idOrSlug)
 }
 
+// GetWithParams retrieves details for an organization.
+// The organization can be fetched by either the ID or its slug.
+func GetWithParams(ctx context.Context, idOrSlug string, params *GetParams) (*clerk.Organization, error) {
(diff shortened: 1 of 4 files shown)


Nikpolik (approved):

nikospapcom (approved):

organization/api.go
```diff
@@ -19,6 +19,12 @@ func Get(ctx context.Context, idOrSlug string) (*clerk.Organization, error) {
 	return getClient().Get(ctx, idOrSlug)
 }
 
+// GetWithParams retrieves details for an organization.
+// The organization can be fetched by either the ID or its slug.
+func GetWithParams(ctx context.Context, idOrSlug string, params *GetParams) (*clerk.Organization, error) {
```
chanioxaris, line 24: ❓ Seems like unrelated changes to the PR description

chanioxaris (approved):

organization/api.go
octoper: I ran `go generate ./...` so this was probably forgotten to run in a previous PR

PR #379, opened 2025-02-12
feat: Add `DisableInvitations` on organization instance settings

izaaklauer:
## What changed?

We're introducing a new property on organization settings to disable invitations via fapi if the application is in restricted or waitlist mode, and this introduces it in the SDK!

## Where are the tests?

It doesn't look like we have property-level test coverage for org settings yet, and the test coverage of this is solid in our internal use of this client, so i'm not worried about it.

### DIFF
--- a/instance_settings.go
+++ b/instance_settings.go
@@ -22,4 +22,5 @@ type OrganizationSettings struct {
 	DomainsEnabled         bool     `json:"domains_enabled"`
 	DomainsEnrollmentModes []string `json:"domains_enrollment_modes"`
 	DomainsDefaultRole     string   `json:"domains_default_role"`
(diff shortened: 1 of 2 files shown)


BrandonRomano (changes requested):
@izaaklauer probably already see it, but I think you want to cherry-pick just the single commit into v2, rather than merging all of v3 + your commit into v2.

izaaklauer:
Ah yup I see it, thank you @BrandonRomano! Wrong target for this branch.

izaaklauer:
New plan:
- Merge the clerk backend changes that this depends on
- Make and merge a new PR here, pointing at the v2 branch
- Cherry pick or merge commit the change to the v3 branch

PR #381, opened 2025-03-06
feat: Add support for session reverification

kostaspt:
This PR adds Go helpers and middleware for session reverification. This allows to do checks based on the factor ages and triggering reverification when needed.

### DIFF
--- a/http/middleware_test.go
+++ b/http/middleware_test.go
@@ -1,6 +1,7 @@
 package http
 
 import (
+	"encoding/json"
 	"fmt"
 	"net/http"
 	"net/http/httptest"
@@ -365,7 +366,7 @@ func TestNeedsSessionReverification(t *testing.T) {
 				})
 			}
 
-			// This is the user's server, guarded by Clerk's middleware.
+			// This is the user's server, using the NeedsSessionReverification middleware.
 			ts := httptest.NewServer(includeSessionClaims(NeedsSessionReverification(tc.policy)(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 				_, err := w.Write([]byte("{}"))
 				require.NoError(t, err)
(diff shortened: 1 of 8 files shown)


http/middleware_test.go
```diff
@@ -283,3 +283,105 @@ func TestAuthorizationJWTExtractor(t *testing.T) {
...
+			req, err := http.NewRequest(http.MethodGet, ts.URL, nil)
+			require.NoError(t, err)
+
+			// Send the request
+			resp, err := ts.Client().Do(req)
+			require.NoError(t, err)
+			defer resp.Body.Close()
+
+			// Verify the response status
+			require.Equal(t, tc.expectedStatus, resp.StatusCode)
```
chanioxaris, line 385: 🔧 We should also assert the response body, as it's critical in order the FE SDK to handle such case. If the response changes by mistake, then the flow is gonna break

jwt.go
```diff
@@ -69,6 +69,34 @@ func (s *SessionClaims) HasRole(role string) bool {
...
+			return firstFactorNeedsReverification
+		}
+		return secondFactorNeedsReverification
+	case SessionReverificationLevelMultiFactor:
+		if !isSecondFactorEnabled {
+			return firstFactorNeedsReverification
+		}
+		return firstFactorNeedsReverification || secondFactorNeedsReverification
+	default:
+		return true
```
chanioxaris, line 96: 💯

chanioxaris (approved):

http/middleware_test.go
kostaspt: Fixed in [`3db743a` (#381)](https://github.com/clerk/clerk-sdk-go/pull/381/commits/3db743ae07a2b57b56299c6bddea2bd6c9f8d096)

PR #416, opened 2025-07-18
adding machine scopes + machine secret to sdk (v2)

Ben2W:
# Machine Scopes, Secret Key, and Default Token TTL Support

This PR adds to the machine functionality in our Go SDK with three major features:

### 🔗 Machine Scopes
Machines can now have access to other machines through scoped relationships:

```go
type MachineWithScopedMachines struct {
    APIResource
    Machine
    ScopedMachines []*Machine `json:"scoped_machines"`
}
```

- Added `ScopedMachines` field to machine responses (array of machine IDs)
- Scoped machines never include secret keys to maintain security

### �� Secret Key Support
Machines now return a `SecretKey` field when created:

```go
type MachineWithScopedMachinesAndSecretKey struct {
    APIResource
    MachineWithScopedMachines
    SecretKey string `json:"secret_key"`
}
```

- Secret key is only included in create responses
- Not exposed in subsequent GET/UPDATE/LIST operations
- Maintains security by not exposing secret keys in scoped machine relationships

###  Default Token TTL
Added `DefaultTokenTTL` field to control token expiration:

```go
type CreateParams struct {
    clerk.APIParams
    Name           string   `json:"name"` // Required for create
    ScopedMachines []string `json:"scoped_machines,omitempty"`
    DefaultTokenTTL *int64  `json:"default_token_ttl,omitempty"`
}
```

- Available as both request parameter (create/update) and response field
- Optional parameter with `omitempty` JSON tag

### Create Machine Name Parameter
```go
// Before
CreateParams{Name: &stringPointer}

// After  
CreateParams{Name: stringValue}
```

Since the Name _was required_ when creating machines, this was invalid. We also doubt anybody was using this function as it wasn't documented


BrandonRomano (approved):
Rare circumstance where this breaking change is OK in the SDK, given this feature is in private beta!

PR #419, opened 2025-07-21
feat(oauth): add `OAuthApplication.ClientURL` & `.ClientImageURL`

jfoshee:
Adds members `ClientURL` and `ClientImageURL` to `OAuthApplication` DTO

These correspond to OAuth Client Metadata from [RFC 7591](https://datatracker.ietf.org/doc/html/rfc7591#:~:text=client_uri%0A%20%20%20%20%20%20URL%20string%20of,in%0A%20%20%20%20%20%20Section%202.2.)
```text
   client_uri
      URL string of a web page providing information about the client.
      If present, the server SHOULD display this URL to the end-user in
      a clickable fashion.  It is RECOMMENDED that clients always send
      this field.  The value of this field MUST point to a valid web
      page.  The value of this field MAY be internationalized...

   logo_uri
      URL string that references a logo for the client.  If present, the
      server SHOULD display this image to the end-user during approval.
      The value of this field MUST point to a valid image file.  The
      value of this field MAY be internationalized...
```

Note that the ClientImageURL that we return from the APIs will be Clerk-hosted.

API Implementation: https://github.com/clerk/clerk_go/pull/12903

Additive-only optional values, so we can back-port to v2.

Part of USER-2011


jeremy-clerk (approved):
Symbolic review but 👍

Ben2W (approved):

PR #421, opened 2025-07-28
feat: Add the ability to toggle allow/block lists on sign in

tmilewski:
Enables the ability to toggle on/of the allowlist and blocklists for sign-in.


BrandonRomano (approved):
You will also want to backport this commit to v2 once this is all live.  Just add that to your list of things so we don't forget.

PR #430, opened 2025-08-07
Added missing parameters for organizationinvitation List request

seeruk:
The documentation for [listing organization invitations](https://clerk.com/docs/reference/backend-api/tag/organization-invitations/get/organizations/%7Borganization_id%7D/invitations) specifies that you can filter by an email address, and you can specify an ordering. These are both missing from the client params type. This PR adds them in!


BrandonRomano (approved):
Thank you for this contribution 🙇

PR #433, opened 2025-08-13
Switch `secret` -> `token` in m2m (v2)

BrandonRomano:
This is a breaking change, but this feature is in private beta, and nobody in the beta is using the go-sdk functionality yet.


wobsoriano (approved):
🧹

PR #438, opened 2025-08-28
feat: Invite or reject waitlist entries

dmoerner:
Add support for Backend API endpoints to invite or reject waitlist entries.


BrandonRomano (approved):
ez!

PR #443, opened 2025-09-09
Add support for Billing endpoints

mauricioabreu:
* List plans
* Extend Free Trials
* List Subscription Items
* Cancel Subscription Item


billing.go
```diff
@@ -0,0 +1,179 @@
...
+	Object      string `json:"object"`
+	ID          string `json:"id"`
+	Name        string `json:"name"`
+	Description string `json:"description"`
+	Slug        string `json:"slug"`
+	AvatarURL   string `json:"avatar_url"`
+}
+
+// CommerceMoney represents money amounts with formatting.
+type CommerceMoney struct {
```
NicolasLopes7: 🔧 Should we be using commerce externally? Haven't we migrated to billing?

billing.go
```diff
@@ -0,0 +1,179 @@
...
+	APIResource
+
+	Amount          int64  `json:"amount"`
+	AmountFormatted string `json:"amount_formatted"`
+	Currency        string `json:"currency"`
+	CurrencySymbol  string `json:"currency_symbol"`
+}
+
+// CommerceProduct represents a product.
+type CommerceProduct struct {
```
NicolasLopes7: 🔧 Should we be using commerce externally? Haven't we migrated to billing?

billing.go
```diff
@@ -0,0 +1,179 @@
...
+// PlanList contains a list of plans.
+type PlanList struct {
+	APIResource
+
+	Data       []*Plan `json:"data"`
+	TotalCount int64   `json:"total_count"`
+}
+
+// CommercePaymentSource represents a payment source.
+type CommercePaymentSource struct {
```
NicolasLopes7: 🔧 Should we be using commerce externally? Haven't we migrated to billing?

billing.go
```diff
@@ -0,0 +1,179 @@
...
+	CardType                 string  `json:"card_type"`
+	ExpiryYear               int     `json:"expiry_year,omitempty"`
+	ExpiryMonth              int     `json:"expiry_month,omitempty"`
+	CreatedAt                int64   `json:"created_at"`
+	UpdatedAt                int64   `json:"updated_at"`
+	IsRemovable              *bool   `json:"is_removable,omitempty"`
+}
+
+// CommerceSubscriptionItemNextPayment represents next payment info.
+type CommerceSubscriptionItemNextPayment struct {
```
NicolasLopes7: 🔧 Should we be using commerce externally? Haven't we migrated to billing?

NicolasLopes7:
You also probably want to open this PR pointing to the `v3` branch (plus v2)

billing.go
mauricioabreu: I'm still not used to it :( 
but yes!

mauricioabreu: Fixed via 40c7cc4e9fb098a0d80031c9b65a9a4887635a2b

billing.go
```diff
@@ -0,0 +1,179 @@
...
+
+// SubscriptionItem represents a billing subscription item.
+type SubscriptionItem struct {
+	APIResource
+
+	Object          string                              `json:"object"`
+	ID              string                              `json:"id"`
+	InstanceID      string                              `json:"instance_id"`
+	Status          string                              `json:"status"`
+	PlanID          string                              `json:"plan_id"`
```
paddycarver: Could we make this a pointer to support the upcoming planless subscription items?

mauricioabreu: Sure!

Solved via 22831dac04d82b3e5544b32f0abadb30c4c27194

billing.go
```diff
@@ -0,0 +1,179 @@
...
+	Name                         string          `json:"name"`
+	Fee                          *BillingMoney   `json:"fee"`
+	AnnualMonthlyFee             *BillingMoney   `json:"annual_monthly_fee"`
+	AnnualFee                    *BillingMoney   `json:"annual_fee"`
+	Amount                       *int64          `json:"amount,omitempty"`
+	AmountFormatted              *string         `json:"amount_formatted,omitempty"`
+	AnnualMonthlyAmount          *int64          `json:"annual_monthly_amount,omitempty"`
+	AnnualMonthlyAmountFormatted *string         `json:"annual_monthly_amount_formatted,omitempty"`
+	AnnualAmount                 *int64          `json:"annual_amount,omitempty"`
+	AnnualAmountFormatted        *string         `json:"annual_amount_formatted,omitempty"`
```
paddycarver: Aren't these deprecated? Could we just... not introduce them? Then we don't need to remove them later.

billing.go
```diff
@@ -0,0 +1,179 @@
...
+	Currency                     string          `json:"currency"`
+	Description                  string          `json:"description"`
+	ProductID                    string          `json:"product_id"`
+	Product                      *BillingProduct `json:"product,omitempty"`
+	IsDefault                    bool            `json:"is_default"`
+	IsRecurring                  bool            `json:"is_recurring"`
+	PubliclyVisible              bool            `json:"publicly_visible"`
+	HasBaseFee                   bool            `json:"has_base_fee"`
+	PayerType                    []string        `json:"payer_type"`
+	ForPayerType                 string          `json:"for_payer_type"`
```
paddycarver: See BILL-1275; do we want both of these?

billing.go
```diff
@@ -0,0 +1,179 @@
...
+	GatewayExternalID        string  `json:"gateway_external_id"`
+	GatewayExternalAccountID *string `json:"gateway_external_account_id"`
+	Last4                    string  `json:"last4"`
+	Status                   string  `json:"status"`
+	WalletType               string  `json:"wallet_type"`
+	CardType                 string  `json:"card_type"`
+	ExpiryYear               int     `json:"expiry_year,omitempty"`
+	ExpiryMonth              int     `json:"expiry_month,omitempty"`
+	CreatedAt                int64   `json:"created_at"`
+	UpdatedAt                int64   `json:"updated_at"`
```
paddycarver, line 90: Possibly unpopular opinion, but should these be `time.Time`s, along with any other date or timestamp field?

billing.go
```diff
@@ -0,0 +1,179 @@
...
+
+	Object     string `json:"object"`
+	ID         string `json:"id"`
+	InstanceID string `json:"instance_id"`
+
+	// User payer only
+	UserID    string `json:"user_id,omitempty"`
+	FirstName string `json:"first_name"`
+	LastName  string `json:"last_name"`
+	Email     string `json:"email"`
```
paddycarver: These should probably be all omitempty, too, right? If we make a request for an org payer, we don't want to include empty values for these, right?

billing.go
```diff
@@ -0,0 +1,179 @@
...
+
+	Object          string                              `json:"object"`
+	ID              string                              `json:"id"`
+	InstanceID      string                              `json:"instance_id"`
+	Status          string                              `json:"status"`
+	PlanID          *string                             `json:"plan_id"`
+	Plan            *Plan                               `json:"plan"`
+	PlanPeriod      string                              `json:"plan_period"`
+	PaymentSourceID string                              `json:"payment_source_id"`
+	PaymentSource   *BillingPaymentSource               `json:"payment_source"`
```
paddycarver: I think we want these to be PaymentMethod, right?

billing.go
```diff
@@ -0,0 +1,179 @@
...
+	LifetimePaid    *BillingMoney                       `json:"lifetime_paid"`
+	Amount          *BillingMoney                       `json:"amount"`
+	NextInvoice     *BillingSubscriptionItemNextPayment `json:"next_invoice"`
```
paddycarver: I'd say let's just not include this, then we don't need to get rid of it, right?

billing.go
```diff
@@ -0,0 +1,179 @@
...
+	Data       []*SubscriptionItem `json:"data"`
+	TotalCount int64               `json:"total_count"`
+}
+
+// ExtendFreeTrialResponse represents the response from extending free trial.
+type ExtendFreeTrialResponse struct {
+	APIResource
+
+	*SubscriptionItem
+}
```
paddycarver: 🤔 should this be a separate type? And if so, should the non-pointer embed the pointer? That leads to some weird stuff where `ExtendFreeTrialResponse` is non-`nil` but the `*SubscriptionItem` is nil.

billing/client.go
```diff
@@ -0,0 +1,162 @@
...
+type ListSubscriptionItemsParams struct {
+	clerk.APIParams
+	clerk.ListParams
+	Status         *string `json:"status,omitempty"`
+	PayerType      *string `json:"payer_type,omitempty"`
+	PlanID         *string `json:"plan_id,omitempty"`
+	IncludeFree    *bool   `json:"include_free,omitempty"`
+	Query          *string `json:"query,omitempty"`
+	UserID         *string `json:"user_id,omitempty"`
+	OrganizationID *string `json:"organization_id,omitempty"`
```
paddycarver: Do these need JSON tags if we're only including them in the URL query params?

billing.go
paddycarver: Wait, we're never actually marshalling these structs. None of these should be omitempty, I think?

paddycarver (changes requested):
I'm assuming those tests are following the pattern for other tests in this repo?

mauricioabreu:
@paddycarver 
> I'm assuming those tests are following the pattern for other tests in this repo?

Yes, they are.

billing.go
mauricioabreu: Yes! I opened this PR before we decided to rename it.

billing.go
mauricioabreu: No, we dont. Also, we are not using it. Removing 🧹

billing.go
mauricioabreu: Solved via 3cb4847c8a67e2c909d0d135628ebdb77bbeea84

billing.go
mauricioabreu: Since this is Unix Time in milliseconds, I think it should be int64 🤔

billing.go
mauricioabreu: Good catch! Solved via 88537fa0b1ae50d3abbe184096058dcfcead63c3

billing.go
mauricioabreu: Solved via 51c15a7c121ff51b3b81bb101d8c3cf597b1e13d

billing/client.go
mauricioabreu: For this struct, I tried to stay consistent with what we already have in `clerk-sdk-go`

billing.go
mauricioabreu: You are right! Removed all `omitempty` from our response structs

mauricioabreu: Solved via f26abcb1a03458d1a5ba3fb0ee49626a2269bd6d

billing.go
paddycarver: Yeah, but we're surfacing it to Go developers. Unix time in milliseconds is the encoding scheme, but the underlying thing it's representing is a timestamp, and in Go that's a `time.Time`.

paddycarver: We decided to do this across the SDK as a whole if that's the direction we want to go in. A problem for another PR. For now, it's better to be consistent.

billing/client.go
```diff
@@ -0,0 +1,162 @@
...
+}
+
+type ListPlansParams struct {
+	clerk.APIParams
+	clerk.ListParams
+	PayerType *string `json:"payer_type,omitempty"`
+}
+
+func (params *ListPlansParams) ToQuery() url.Values {
+	q := params.ListParams.ToQuery()
```
paddycarver, line 34: Does this need `params.APIParams.ToQuery()`, too?

billing/client.go
```diff
@@ -0,0 +1,162 @@
...
+func NewClient(config *clerk.ClientConfig) *Client {
+	return &Client{
+		Backend: clerk.NewBackend(&config.BackendConfig),
+	}
+}
+
+type ListPlansParams struct {
+	clerk.APIParams
+	clerk.ListParams
+	PayerType *string `json:"payer_type,omitempty"`
```
paddycarver: Does this need a JSON tag?

billing/client.go
```diff
@@ -0,0 +1,162 @@
...
+	PayerType      *string `json:"payer_type"`
+	PlanID         *string `json:"plan_id"`
+	IncludeFree    *bool   `json:"include_free"`
+	Query          *string `json:"query"`
+	UserID         *string `json:"user_id"`
+	OrganizationID *string `json:"organization_id"`
+}
+
+func (params *ListSubscriptionItemsParams) ToQuery() url.Values {
+	q := params.ListParams.ToQuery()
```
paddycarver, line 67: Does this need a `params.APIParams.ToQuery()`?

billing/client.go
```diff
@@ -0,0 +1,162 @@
...
+type ListSubscriptionItemsParams struct {
+	clerk.APIParams
+	clerk.ListParams
+	Status         *string `json:"status"`
+	PayerType      *string `json:"payer_type"`
+	PlanID         *string `json:"plan_id"`
+	IncludeFree    *bool   `json:"include_free"`
+	Query          *string `json:"query"`
+	UserID         *string `json:"user_id"`
+	OrganizationID *string `json:"organization_id"`
```
paddycarver: Do these all need JSON tags?

billing/client.go
```diff
@@ -0,0 +1,162 @@
...
+	return subscriptionItemList, err
+}
+
+type CancelSubscriptionItemParams struct {
+	clerk.APIParams
+	EndNow *bool `json:"end_now,omitempty"`
+}
+
+func (params *CancelSubscriptionItemParams) ToQuery() url.Values {
+	q := url.Values{}
```
paddycarver, line 115: Does this need an `APIParams.ToQuery()`?

billing/client.go
```diff
@@ -0,0 +1,162 @@
...
+	req := clerk.NewAPIRequest(http.MethodDelete, cancelPath)
+	subscriptionItem := &clerk.SubscriptionItem{}
+	err = c.Backend.Call(ctx, req, subscriptionItem)
+	return subscriptionItem, err
+}
+
+type ExtendFreeTrialParams struct {
+	clerk.APIParams
+	ExtendTo string `json:"extend_to"`
+}
```
paddycarver, line 149: These don't appear to implement the `ToQuery` method, unlike all the others? Why not?

billing.go
```diff
@@ -0,0 +1,165 @@
...
+
+	Object     string `json:"object"`
+	ID         string `json:"id"`
+	InstanceID string `json:"instance_id"`
+
+	// User payer only
+	UserID    string `json:"user_id"`
+	FirstName string `json:"first_name"`
+	LastName  string `json:"last_name"`
+	Email     string `json:"email"`
```
paddycarver: Should these be pointers? They can be omitted/set to null, right? And we may want to distinguish that from `""`?

billing.go
```diff
@@ -0,0 +1,165 @@
...
+
+	// Org payer only
+	OrganizationID   string `json:"organization_id"`
+	OrganizationName string `json:"organization_name"`
```
paddycarver: Same here, should these be pointers?

billing.go
```diff
@@ -0,0 +1,165 @@
...
+	APIResource
+
+	Object           string          `json:"object"`
+	ID               string          `json:"id"`
+	Name             string          `json:"name"`
+	Fee              *BillingMoney   `json:"fee"`
+	AnnualMonthlyFee *BillingMoney   `json:"annual_monthly_fee"`
+	AnnualFee        *BillingMoney   `json:"annual_fee"`
+	CurrencySymbol   string          `json:"currency_symbol"`
+	Currency         string          `json:"currency"`
```
paddycarver: Are the top level Currency and CurrencySymbol still used?

billing.go
```diff
@@ -0,0 +1,165 @@
...
+type BillingProduct struct {
+	APIResource
+
+	Object    string  `json:"object"`
+	ID        string  `json:"id"`
+	Slug      string  `json:"slug"`
+	Currency  string  `json:"currency"`
+	Name      string  `json:"name"`
+	IsDefault bool    `json:"is_default"`
+	Plans     []*Plan `json:"plans"`
```
paddycarver:
```suggestion
	Plans     []Plan `json:"plans"`
```

I don't think we'll ever have a nil plan in this list, right?

billing.go
```diff
@@ -0,0 +1,165 @@
...
+	CurrencySymbol   string          `json:"currency_symbol"`
+	Currency         string          `json:"currency"`
+	Description      string          `json:"description"`
+	ProductID        string          `json:"product_id"`
+	Product          *BillingProduct `json:"product"`
+	IsDefault        bool            `json:"is_default"`
+	IsRecurring      bool            `json:"is_recurring"`
+	PubliclyVisible  bool            `json:"publicly_visible"`
+	HasBaseFee       bool            `json:"has_base_fee"`
+	PayerType        []string        `json:"payer_type"`
```
paddycarver: Should we only support ForPayerType, like https://github.com/clerk/javascript/pull/6342 ?

billing.go
```diff
@@ -0,0 +1,165 @@
...
+	ForPayerType     string          `json:"for_payer_type"`
+	Slug             string          `json:"slug"`
+	AvatarURL        string          `json:"avatar_url"`
+	Features         []*Feature      `json:"features"`
```
paddycarver:
```suggestion
	Features         []Feature      `json:"features"`
```

We're not returning a nil feature, right?

billing.go
```diff
@@ -0,0 +1,165 @@
...
+}
+
+// Plan represents a billing plan.
+type Plan struct {
+	APIResource
+
+	Object           string          `json:"object"`
+	ID               string          `json:"id"`
+	Name             string          `json:"name"`
+	Fee              *BillingMoney   `json:"fee"`
```
paddycarver, line 45: Do we always have a Fee and only sometimes have an annual fee? Or are there times where we may not have a fee, like the default free plan? 🤔

billing.go
```diff
@@ -0,0 +1,165 @@
...
+	Features         []*Feature      `json:"features"`
+	FreeTrialEnabled bool            `json:"free_trial_enabled"`
+	FreeTrialDays    *int            `json:"free_trial_days"`
+}
+
+// PlanList contains a list of plans.
+type PlanList struct {
+	APIResource
+
+	Data       []*Plan `json:"data"`
```
paddycarver:
```suggestion
	Data       []Plan `json:"data"`
```

We won't return nil plans, right?

billing.go
```diff
@@ -0,0 +1,165 @@
...
+	UpdatedAt                int64   `json:"updated_at"`
+	IsRemovable              *bool   `json:"is_removable"`
+}
+
+// BillingSubscriptionItemNextPayment represents next payment info.
+type BillingSubscriptionItemNextPayment struct {
+	APIResource
+
+	Amount *BillingMoney `json:"amount"`
+	Date   *int64        `json:"date"`
```
paddycarver, line 99: Will these items ever not be set when the outer struct is present?

billing.go
```diff
@@ -0,0 +1,165 @@
...
+	PlanPeriod      string                              `json:"plan_period"`
+	PaymentMethodID string                              `json:"payment_method_id"`
+	PaymentMethod   *BillingPaymentMethod               `json:"payment_method"`
+	LifetimePaid    *BillingMoney                       `json:"lifetime_paid"`
+	Amount          *BillingMoney                       `json:"amount"`
+	NextPayment     *BillingSubscriptionItemNextPayment `json:"next_payment"`
+	PayerID         string                              `json:"payer_id"`
+	Payer           *Payer                              `json:"payer"`
+	IsFreeTrial     bool                                `json:"is_free_trial"`
+	PeriodStart     *int64                              `json:"period_start"`
```
paddycarver: Is it possible for a subscription item to not have a period start?

billing.go
```diff
@@ -0,0 +1,165 @@
...
+	EndedAt         *int64                              `json:"ended_at"`
+	CreatedAt       int64                               `json:"created_at"`
+	UpdatedAt       int64                               `json:"updated_at"`
+}
+
+// SubscriptionItemList contains a list of subscription items.
+type SubscriptionItemList struct {
+	APIResource
+
+	Data       []*SubscriptionItem `json:"data"`
```
paddycarver:
```suggestion
	Data       []SubscriptionItem `json:"data"`
```

We won't include nil subscription items in responses, right?

paddycarver (approved):
Type system nitpicks, feel free to use your judgment on whether to block on them or not.

PR #450, opened 2025-09-22
feat: Add RoleSet resource and Organization integration

NicolasLopes7:
Adds complete RoleSet API functionality with CRUD operations, role management, and Organization RoleSetKey field integration. It also re-runs `go generate ./...` and creates an api definition for the user's API that was missing. `ListWithoutCount`.


LauraBeatris (approved):

user/api.go
```diff
@@ -49,6 +49,12 @@ func List(ctx context.Context, params *ListParams) (*clerk.UserList, error) {
 	return getClient().List(ctx, params)
 }
 
+// ListWithoutCount returns a list of users without fetching the total count.
+// This is more efficient than List when you don't need the total count.
+func ListWithoutCount(ctx context.Context, params *ListParams) ([]*clerk.User, error) {
+	return getClient().ListWithoutCount(ctx, params)
+}
```
iagodahlem, line 56: ❓ Is this change expected? 🤔

iagodahlem (approved):

user/api.go
NicolasLopes7: yup 

>  It also re-runs `go generate ./...` and creates an api definition for the user's API that was missing. `ListWithoutCount`.

added on the pr description

iagodahlem: Oh, nice! Good catch!

mauricioabreu (approved):

PR #451, opened 2025-09-22
feat: Add RoleSet resource and Organization integration (v2)

NicolasLopes7:
Adds complete RoleSet API functionality with CRUD operations, role management, and Organization RoleSetKey field integration. It also re-runs `go generate ./...` and creates an api definition for the user's API that was missing. `ListWithoutCount`.


LauraBeatris (approved):

role_set.go
```diff
@@ -0,0 +1,34 @@
...
+
+// RoleSetItem represents a role within a role set.
+type RoleSetItem struct {
+	Object      string  `json:"object"`
+	ID          string  `json:"id"`
+	Name        string  `json:"name"`
+	Key         string  `json:"key"`
+	Description *string `json:"description"`
+	CreatedAt   int64   `json:"created_at"`
+	UpdatedAt   int64   `json:"updated_at"`
```
iagodahlem, line 26: 🙃 I'm not sure if you are leaving this for another PR, but we should include the "default role for members" boolean here as well.

iagodahlem (approved):

role_set.go
iagodahlem: Just checked that we don't have that column on the DB yet, so all good!

mauricioabreu (approved):

role_set.go
NicolasLopes7: yep yep, will do it later on

PR #461, opened 2025-10-20
feat: add missing User struct fields

kduprey:
## Changes

- Add `EnterpriseAccount` type with SAML/OAuth enterprise connection support
- Enhance `User` struct with new fields:
  - `CreatedAt`, `UpdatedAt`, `MFAEnabledAt`, `MFADisabledAt`, `LegalAcceptedAt`
  - `EnterpriseAccounts` for enterprise SSO support
- Update `EmailAddress` with `MatchesSsoConnection` field
- Add shared `Verification` and `LinkedIdentification` types

### DIFF
--- a/email_address.go
+++ b/email_address.go
@@ -24,6 +24,7 @@ type Verification struct {
 	ExpireAt                        *int64          `json:"expire_at"`
 	VerifiedAtClient                string          `json:"verified_at_client,omitempty"`
 	Nonce                           *string         `json:"nonce,omitempty"`
+	Message                         *string         `json:"message,omitempty"`
 	ExternalVerificationRedirectURL *string         `json:"external_verification_redirect_url,omitempty"`
 	Error                           json.RawMessage `json:"error,omitempty"`
 }
--- a/email_address.go
+++ b/email_address.go
@@ -4,17 +4,22 @@ import "encoding/json"
 
 type EmailAddress struct {
 	APIResource
-	ID           string                  `json:"id"`
-	Object       string                  `json:"object"`
-	EmailAddress string                  `json:"email_address"`
-	Reserved     bool                    `json:"reserved"`
-	Verification *Verification           `json:"verification"`
-	LinkedTo     []*LinkedIdentification `json:"linked_to"`
+	ID                   string                  `json:"id"`
+	Object               string                  `json:"object"`
+	EmailAddress         string                  `json:"email_address"`
+	Reserved             bool                    `json:"reserved"`
+	Verification         *Verification           `json:"verification"`
+	LinkedTo             []*LinkedIdentification `json:"linked_to"`
+	MatchesSsoConnection bool                    `json:"matches_sso_connection"`
+	CreatedAt            int64                   `json:"created_at"`
+	UpdatedAt            int64                   `json:"updated_at"`
 }
 
 type Verification struct {
+	Object                          string          `json:"object"`
 	Status                          string          `json:"status"`
 	Strategy                        string          `json:"strategy"`
+	Channel                         *string         `json:"channel,omitempty"`
 	Attempts                        *int64          `json:"attempts"`
 	ExpireAt                        *int64          `json:"expire_at"`
 	VerifiedAtClient                string          `json:"verified_at_client,omitempty"`
--- a/email_address.go
+++ b/email_address.go
@@ -10,7 +10,7 @@ type EmailAddress struct {
 	Reserved             bool                    `json:"reserved"`
 	Verification         *Verification           `json:"verification"`
 	LinkedTo             []*LinkedIdentification `json:"linked_to"`
-	MatchesSsoConnection bool                    `json:"matches_sso_connection"`
+	MatchesSSOConnection bool                    `json:"matches_sso_connection"`
 	CreatedAt            int64                   `json:"created_at"`
 	UpdatedAt            int64                   `json:"updated_at"`
 }
--- a/enterprise_account.go
+++ b/enterprise_account.go
@@ -1,0 +1,44 @@
+package clerk
+
+import "encoding/json"
+
+type EnterpriseAccount struct {
+	ID                     string                       `json:"id"`
+	Object                 string                       `json:"object"`
+	Protocol               string                       `json:"protocol"`
+	Provider               string                       `json:"provider"`
+	Active                 bool                         `json:"active"`
+	EmailAddress           string                       `json:"email_address"`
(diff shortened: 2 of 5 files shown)


email_address.go
```diff
@@ -4,17 +4,22 @@ import "encoding/json"
...
-	Reserved     bool                    `json:"reserved"`
-	Verification *Verification           `json:"verification"`
-	LinkedTo     []*LinkedIdentification `json:"linked_to"`
+	ID                   string                  `json:"id"`
+	Object               string                  `json:"object"`
+	EmailAddress         string                  `json:"email_address"`
+	Reserved             bool                    `json:"reserved"`
+	Verification         *Verification           `json:"verification"`
+	LinkedTo             []*LinkedIdentification `json:"linked_to"`
+	MatchesSsoConnection bool                    `json:"matches_sso_connection"`
```
jfoshee: Should this be `MatchesSSOConnection` or do we have an existing style exception for this?

enterprise_account.go
```diff
@@ -0,0 +1,44 @@
...
+	ID                     string `json:"id"`
+	EnterpriseConnectionID string `json:"enterprise_connection_id"`
+	Protocol               string `json:"protocol"`
+	Provider               string `json:"provider"`
+
+	// Name is the name of this enterprise connection that we will display directly to end-users
+	Name          string  `json:"name"`
+	LogoPublicURL *string `json:"logo_public_url"`
+	// Deprecated: Use Domains instead
+	Domain                           string   `json:"domain,omitempty"`
```
jfoshee: ❓  Honest question: as `domain` is already deprecated and isn't already a part of the Go SDK, do we need to include it here?

enterprise_account.go
```diff
@@ -0,0 +1,44 @@
...
+	Domains                          []string `json:"domains,omitempty"`
+	Active                           bool     `json:"active"`
+	SyncUserAttributes               bool     `json:"sync_user_attributes"`
+	DisableAdditionalIdentifications bool     `json:"disable_additional_identifications"`
+	CreatedAt                        int64    `json:"created_at"`
+	UpdatedAt                        int64    `json:"updated_at"`
+
+	// SAML Connection specific fields
+	AllowSubdomains   bool `json:"allow_subdomains"`
+	AllowIdpInitiated bool `json:"allow_idp_initiated"`
```
jfoshee: Similar to above, should we make this `AllowIDPInitiated`?

user.go
```diff
@@ -66,23 +71,44 @@ type Web3Wallet struct {
...
+	ID                     string                 `json:"id"`
+	Provider               string                 `json:"provider"`
+	Active                 bool                   `json:"active"`
+	EmailAddress           string                 `json:"email_address"`
+	FirstName              *string                `json:"first_name"`
+	LastName               *string                `json:"last_name"`
+	ProviderUserID         *string                `json:"provider_user_id"`
+	LastAuthenticatedAt    *int64                 `json:"last_authenticated_at"`
+	PublicMetadata         json.RawMessage        `json:"public_metadata"`
+	SamlConnection         *samlAccountConnection `json:"saml_connection"`
```
jfoshee: I guess it would be a breaking change to call this type `SAMLConnection`? :-/

user.go
```diff
@@ -66,23 +71,44 @@ type Web3Wallet struct {
...
 	APIResource
 	Users      []*User `json:"data"`
 	TotalCount int64   `json:"total_count"`
 }
+
+type samlAccountConnection struct {
+	ID   string `json:"id"`
+	Name string `json:"name"`
+	// Deprecated: Use Domains instead
+	Domain                           string   `json:"domain"`
```
jfoshee: Should `domain` be included, as it is already deprecated?

user.go
```diff
@@ -66,23 +71,44 @@ type Web3Wallet struct {
...
+	Domains                          []string `json:"domains"`
+	Active                           bool     `json:"active"`
+	Provider                         string   `json:"provider"`
+	SyncUserAttributes               bool     `json:"sync_user_attributes"`
+	AllowSubdomains                  bool     `json:"allow_subdomains"`
+	AllowIdpInitiated                bool     `json:"allow_idp_initiated"`
```
jfoshee: Should `IDP` be capitalized?

email_address.go
kduprey: This is currently the same name as what is in our [main Go repo](https://github.com/clerk/clerk_go/blob/kenton/user-2719-add-user-object-to-sessioncreated-webhook/api/serialize/identification.go#L19-L20).

enterprise_account.go
kduprey: This is the same name as what is in our [main Go repo](https://github.com/clerk/clerk_go/blob/kenton/user-2719-add-user-object-to-sessioncreated-webhook/api/serialize/enterprise_account.go#L54-L55).

user.go
kduprey: This is currently the same name as what is in our [main Go repo](https://github.com/clerk/clerk_go/blob/kenton/user-2719-add-user-object-to-sessioncreated-webhook/api/serialize/saml_account.go#L36-L37).

user.go
kduprey: Yes, due to the change in name. I will still call out that this is the same name as what is in our [main Go repo](https://github.com/clerk/clerk_go/blob/kenton/user-2719-add-user-object-to-sessioncreated-webhook/api/serialize/saml_account.go#L21-L22).

email_address.go
jfoshee: Granted. But I'm not convinced that because we are not following Go conventions in our internal repo, that we should let that unduly influence our public Go repo. If we decide to rename this later it becomes a breaking change for our customers. But internally we can usually rename more easily. So I think we should try to "get it right the first time" in this repo.

enterprise_account.go
kduprey: Removed

user.go
kduprey: Removed

user.go
kduprey: Wait, scratch that. We can change the name without breaking change as the field was not present previously.

enterprise_account.go
kduprey: Updated

user.go
kduprey: Updated

user.go
kduprey: Updated

jfoshee (approved):
Thanks!

jfoshee (approved):
LGTM, Thanks

PR #465, opened 2025-11-04
feat: Updating sdk to use the new API version

l-armstrong:
We want to update the sdk to use the new API version. This PR also contains updates to some of the billing types reflected in the new API version.

### DIFF
--- a/billing.go
+++ b/billing.go
@@ -12,8 +12,8 @@ type Feature struct {
 	AvatarURL   string `json:"avatar_url"`
 }
 
-// BillingMoney represents money amounts with formatting.
-type BillingMoney struct {
+// BillingFee represents money amounts with formatting.
+type BillingFee struct {
 	APIResource
 
 	Amount          int64  `json:"amount"`
@@ -32,32 +32,31 @@ type BillingProduct struct {
 	Currency  string `json:"currency"`
 	Name      string `json:"name"`
 	IsDefault bool   `json:"is_default"`
-	Plans     []Plan `json:"plans"`
+	Plans     []Plan `json:"plans,omitempty"`
 }
 
 // Plan represents a billing plan.
 type Plan struct {
 	APIResource
 
-	Object           string          `json:"object"`
-	ID               string          `json:"id"`
-	Name             string          `json:"name"`
-	Fee              *BillingMoney   `json:"fee"`
-	AnnualMonthlyFee *BillingMoney   `json:"annual_monthly_fee"`
-	AnnualFee        *BillingMoney   `json:"annual_fee"`
-	Description      string          `json:"description"`
-	ProductID        string          `json:"product_id"`
-	Product          *BillingProduct `json:"product"`
-	IsDefault        bool            `json:"is_default"`
-	IsRecurring      bool            `json:"is_recurring"`
-	PubliclyVisible  bool            `json:"publicly_visible"`
-	HasBaseFee       bool            `json:"has_base_fee"`
-	ForPayerType     string          `json:"for_payer_type"`
-	Slug             string          `json:"slug"`
-	AvatarURL        string          `json:"avatar_url"`
-	Features         []Feature       `json:"features"`
-	FreeTrialEnabled bool            `json:"free_trial_enabled"`
-	FreeTrialDays    *int            `json:"free_trial_days"`
+	Object           string      `json:"object"`
+	ID               string      `json:"id"`
+	Name             string      `json:"name"`
+	Fee              *BillingFee `json:"fee"`
+	AnnualMonthlyFee *BillingFee `json:"annual_monthly_fee"`
+	AnnualFee        *BillingFee `json:"annual_fee"`
+	Description      *string     `json:"description"`
+	ProductID        string      `json:"product_id"`
+	IsDefault        bool        `json:"is_default"`
+	IsRecurring      bool        `json:"is_recurring"`
+	PubliclyVisible  bool        `json:"publicly_visible"`
+	HasBaseFee       bool        `json:"has_base_fee"`
+	ForPayerType     string      `json:"for_payer_type"`
+	Slug             string      `json:"slug"`
+	AvatarURL        *string     `json:"avatar_url"`
+	Features         []Feature   `json:"features,omitempty"`
+	FreeTrialEnabled bool        `json:"free_trial_enabled"`
+	FreeTrialDays    *int        `json:"free_trial_days"`
 }
 
 // PlanList contains a list of plans.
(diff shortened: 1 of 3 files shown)


billing.go
```diff
@@ -32,32 +32,31 @@ type BillingProduct struct {
 	Currency  string `json:"currency"`
 	Name      string `json:"name"`
 	IsDefault bool   `json:"is_default"`
-	Plans     []Plan `json:"plans"`
+	Plans     []Plan `json:"plans,omitempty"`
```
paddycarver: This should be unnecessary; that only matters when we're going from Go struct -> JSON. The API responses go from JSON -> Go struct, which the `omitempty` tag doesn't impact.

billing.go
```diff
@@ -32,32 +32,31 @@ type BillingProduct struct {
...
+	Description      *string     `json:"description"`
+	ProductID        string      `json:"product_id"`
+	IsDefault        bool        `json:"is_default"`
+	IsRecurring      bool        `json:"is_recurring"`
+	PubliclyVisible  bool        `json:"publicly_visible"`
+	HasBaseFee       bool        `json:"has_base_fee"`
+	ForPayerType     string      `json:"for_payer_type"`
+	Slug             string      `json:"slug"`
+	AvatarURL        *string     `json:"avatar_url"`
+	Features         []Feature   `json:"features,omitempty"`
```
paddycarver: Unnecessary, `omitempty` only impacts serializing, not deserializing.

billing.go
```diff
@@ -75,28 +74,28 @@ type BillingPaymentMethod struct {
...
-	ExpiryMonth              int     `json:"expiry_month"`
-	CreatedAt                int64   `json:"created_at"`
-	UpdatedAt                int64   `json:"updated_at"`
-	IsRemovable              *bool   `json:"is_removable"`
+	WalletType               *string `json:"wallet_type,omitempty"`
+	CardType                 *string `json:"card_type"`
+	ExpiryYear               *int    `json:"expiry_year,omitempty"`
+	ExpiryMonth              *int    `json:"expiry_month,omitempty"`
+	CreatedAt                int64   `json:"created_at,omitempty"`
+	UpdatedAt                int64   `json:"updated_at,omitempty"`
```
paddycarver: Shouldn't need any of these `omitempty`s either.

billing.go
```diff
@@ -12,8 +12,8 @@ type Feature struct {
 	AvatarURL   string `json:"avatar_url"`
 }
 
-// BillingMoney represents money amounts with formatting.
-type BillingMoney struct {
+// BillingFee represents money amounts with formatting.
+type BillingFee struct {
```
paddycarver: Hot take: I actually like this as BillingMoney more than BillingFee; the `fee` property on e.g. plans represents an amount of money, but other things (i.e., credits) could be expressed with the same money type, and they aren't really "fees".

billing.go
```diff
@@ -109,48 +108,48 @@ type Payer struct {
...
+	LastName  *string `json:"last_name,omitempty"`
+	Email     *string `json:"email,omitempty"`
 
 	// Org payer only
 	OrganizationID   *string `json:"organization_id"`
 	OrganizationName *string `json:"organization_name"`
 
 	// Used for both org and user payers
-	ImageURL string `json:"image_url"`
+	ImageURL string `json:"image_url,omitempty"`
```
paddycarver: The omitempty doesn't really do anything for this, but I think making it a pointer would be idiomatic. It would allow people to detect when the ImageURL wasn't set.

billing.go
```diff
@@ -109,48 +108,48 @@ type Payer struct {
...
+	ID              string                             `json:"id"`
+	InstanceID      string                             `json:"instance_id"`
+	Status          string                             `json:"status"`
+	PlanID          *string                            `json:"plan_id"`
+	Plan            *Plan                              `json:"plan,omitempty"`
+	PlanPeriod      string                             `json:"plan_period"`
+	PaymentMethodID string                             `json:"payment_method_id"`
+	PaymentMethod   *BillingPaymentMethod              `json:"payment_method,omitempty"`
+	LifetimePaid    BillingFee                         `json:"lifetime_paid,omitempty"`
+	Amount          BillingFee                         `json:"amount,omitempty"`
```
paddycarver: Same thing here, I'd use pointers instead of omitempty.

billing.go
```diff
@@ -109,48 +108,48 @@ type Payer struct {
...
+	NextPayment     BillingSubscriptionItemNextPayment `json:"next_payment,omitempty"`
+	PayerID         string                             `json:"payer_id"`
+	Payer           Payer                              `json:"payer,omitempty"`
```
paddycarver: I'd use a pointer instead of omitempty.

billing.go
```diff
@@ -109,48 +108,48 @@ type Payer struct {
...
+	InstanceID      string                             `json:"instance_id"`
+	Status          string                             `json:"status"`
+	PlanID          *string                            `json:"plan_id"`
+	Plan            *Plan                              `json:"plan,omitempty"`
+	PlanPeriod      string                             `json:"plan_period"`
+	PaymentMethodID string                             `json:"payment_method_id"`
+	PaymentMethod   *BillingPaymentMethod              `json:"payment_method,omitempty"`
+	LifetimePaid    BillingFee                         `json:"lifetime_paid,omitempty"`
+	Amount          BillingFee                         `json:"amount,omitempty"`
+	NextPayment     BillingSubscriptionItemNextPayment `json:"next_payment,omitempty"`
```
paddycarver: Pointer instead of omitempty.

billing.go
```diff
@@ -109,48 +108,48 @@ type Payer struct {
...
+	PayerID         string                             `json:"payer_id"`
+	Payer           Payer                              `json:"payer,omitempty"`
+	IsFreeTrial     bool                               `json:"is_free_trial"`
+	PeriodStart     int64                              `json:"period_start"`
+	PeriodEnd       *int64                             `json:"period_end"`
+	ProrationDate   string                             `json:"proration_date,omitempty"`
```
paddycarver: Pointer instead of omitempty.

paddycarver (changes requested):
Mostly just tweaks around naming and omitempty. Pointers would be more appropriate in some cases, I think. And even when serializing, omitempty has no effect on non-pointer structs. :(

billing.go
NicolasLopes7: From a fresh eyes/external pov, I also agree, I think billingMoney is more clear as a type to define any number that is money

billing.go
l-armstrong: removed.

billing.go
l-armstrong: removed.

billing.go
l-armstrong: I changed it back to BillingMoney. I agree, BillingMoney is expressed better in certain context than Fee.

mauricioabreu: Yeah, I was the one who named it `BillingMoney`. While I still think it is better, the reason someone might prefer `BillingFee` is that it is consistent with our backend

billing/client_test.go
```diff
@@ -326,3 +326,454 @@ func TestBillingClientListSubscriptionItemsWithMultipleFilters(t *testing.T) {
...
+	t.Parallel()
+
+	jsonData := `{"object":"payment_method","id":"pm_123","payer_id":"payer_456","payment_type":"bank_account","gateway":"stripe","gateway_external_id":"ext_123","last4":"1234","status":"active","is_removable":true}`
+	var paymentMethod clerk.BillingPaymentMethod
+	err := json.Unmarshal([]byte(jsonData), &paymentMethod)
+	require.NoError(t, err)
+	require.Equal(t, "pm_123", paymentMethod.ID)
+	require.Equal(t, "bank_account", paymentMethod.PaymentType)
+
+	// These fields have omitempty and are absent from the JSON response
```
mauricioabreu: ❔ do they have the `omitempty`? I think they are just pointers
Maybe these testes are outdated

mzhong9723 (approved):
@paddycarver may have to give it an approval before merging is unblocked though!

clerk.go
```diff
@@ -21,7 +21,7 @@ import (
 
 const (
 	sdkVersion      string = "v2.0.0"
-	clerkAPIVersion string = "2025-04-10"
+	clerkAPIVersion string = "2025-11-03"
```
paddycarver: Shouldn't this be `2025-11-10`?

l-armstrong: good catch.

paddycarver (approved):

PR #475, opened 2025-11-17
Update OrganizationInvitation to include PublicInviterData

kalafut:
This OrganizationInvitation output was expanded in https://github.com/clerk/clerk_go/pull/14935 to include inviter details. This PR updates the SDK struct to unmarshal the new fields.


BrandonRomano (approved):

organization_invitation.go
```diff
@@ -20,6 +20,7 @@ type OrganizationInvitation struct {
 	OrganizationID         string                  `json:"organization_id"`
 	InviterID              string                  `json:"inviter_id"`
 	PublicOrganizationData *PublicOrganizationData `json:"public_organization_data,omitempty"`
+	PublicInviterData      *PublicInviterData      `json:"public_inviter_data,omitempty"`
```
NicolasLopes7: worth adding a comment about the cases public invite data?

if it's created via the clerk-sdk-go it will probably never show up as it's bapi.

kalafut: Added in e481e2d.

NicolasLopes7 (approved):
Looks good to me! Only addition i'd make is adding an assertion on the returned object from the organization invitation test to make sure that the public invitee data maps nicely

kalafut:
> Looks good to me! Only addition i'd make is adding an assertion on the returned object from the organization invitation test to make sure that the public invitee data maps nicely

Updated tests in d6df33c39.

PR #478, opened 2025-12-01
feat: Add `BypassClientTrust`

tmilewski:
Adds `BypassClientTrust`/`bypass_client_trust` to the following endpoints:

- User.list
- User.get
- User.create
- User.update


BrandonRomano (approved):

BrandonRomano:
I bet we have a test here which is failing as a result of a full response check - might need to modify a test?

Edit: Oh no just looks like a race condition 🤔

PR #480, opened 2025-12-02
feat: Add `PasswordCompromised` endpoint

octoper:
Adds `/password_compromised` endpoint  and adds `requires_password_reset` to `User.list` and `User.get` responses

### DIFF
--- a/user/client.go
+++ b/user/client.go
@@ -589,14 +589,14 @@ func (c *Client) DeleteExternalAccount(ctx context.Context, params *DeleteExtern
 	return resource, err
 }
 
-type PasswordUntrustedParams struct {
+type PasswordCompromisedParams struct {
 	clerk.APIParams
 	UserID string `json:"user_id"`
 }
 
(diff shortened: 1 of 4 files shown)


user/client.go
```diff
@@ -588,3 +588,20 @@ func (c *Client) DeleteExternalAccount(ctx context.Context, params *DeleteExtern
...
 	return resource, err
 }
+
+type PasswordCompromisedParams struct {
+	clerk.APIParams
+	UserID string `json:"user_id"`
+}
+
+// PasswordCompromised marks the user's password as compromised.
+func (c *Client) PasswordCompromised(ctx context.Context, params *PasswordCompromisedParams) (*clerk.User, error) {
```
chanioxaris: 🔧 maybe we don't need a params struct here and we can pass the userID directly as a standalone parameter

```suggestion
func (c *Client) PasswordCompromised(ctx context.Context, userID string) (*clerk.User, error) {
```

chanioxaris (approved):

PR #481, opened 2025-12-03
refactor: drop support to bulk role set role removal

gabrielmeloc22:
### Context
Removing a role set from a role might trigger a reassignment, in order to avoid many role set migrations, we'll for now keep role removal a single operation

### Disclaimer 
Role sets are not out in production, so we can just update the field and not worry about deprecating it first!

### DIFF
--- a/roleset/api.go
+++ b/roleset/api.go
@@ -39,8 +39,8 @@ func AddRoles(ctx context.Context, roleSetKeyOrID string, params *AddRolesParams
 }
 
 // RemoveRoles removes roles from a role set.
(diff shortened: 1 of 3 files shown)


kalafut (approved):
Not specific to this PR exactly... but did we update the OpenAPI too?

gabrielmeloc22:
> Not specific to this PR exactly... but did we update the OpenAPI too?

Not yet, I'll add them to BAPI's PR, even though we're making a lot of changes, I think is for sure a good idea to keep it updated!

PR #485, opened 2025-12-05
fix: Change PasswordCompromised to accept `revoke_all_sessions` param

octoper:
Change the `user.PasswordCompromised` to accept params so that we can use `revoke_all_sessions` param


NicolasLopes7 (approved):
Approving for expediency

Is that available for customers already? If so that would be a breaking change, right?

If not take this as a ✅

octoper:
@NicolasLopes7 No this is not available to customers yet, also v3 is marked as unstable

chanioxaris (approved):

PR #486, opened 2025-12-05
ORGS-1080: Add reassignment mappings support for role set operations

NicolasLopes7:
Add support for reassignment mappings when updating organizations and deleting role sets. This allows mapping old role keys to new role keys during role set transitions. The `ReassignmentMappings` type is a preview feature and not yet available for all customers. Also renames `RemoveRoles` to `RemoveRole` for API consistency and updates the role set delete endpoint to use POST `/replace` method.

### DIFF
--- a/roleset/api.go
+++ b/roleset/api.go
@@ -24,8 +24,8 @@ func Update(ctx context.Context, roleSetKeyOrID string, params *UpdateParams) (*
 }
 
 // Delete removes a role set.
-func Delete(ctx context.Context, roleSetKeyOrID string) (*clerk.DeletedResource, error) {
-	return getClient().Delete(ctx, roleSetKeyOrID)
+func Delete(ctx context.Context, roleSetKeyOrID string, params *DeleteParams) (*clerk.DeletedResource, error) {
+	return getClient().Delete(ctx, roleSetKeyOrID, params)
 }
 
 // List returns a list of role sets.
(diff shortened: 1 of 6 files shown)


roleset/api.go
```diff
@@ -39,7 +39,7 @@ func AddRoles(ctx context.Context, roleSetKeyOrID string, params *AddRolesParams
 }
 
 // RemoveRoles removes roles from a role set.
-func RemoveRoles(ctx context.Context, roleSetKeyOrID string, params *RemoveRoleParams) (*clerk.RoleSet, error) {
+func RemoveRole(ctx context.Context, roleSetKeyOrID string, params *RemoveRoleParams) (*clerk.RoleSet, error) {
```
NicolasLopes7, line 42: Apparently that was missed before

roleset/client.go
```diff
@@ -82,13 +82,20 @@ func (c *Client) Update(ctx context.Context, roleSetKeyOrID string, params *Upda
 	return roleSet, err
 }
 
+type DeleteParams struct {
+	clerk.APIParams
+	ReassignmentMappings *clerk.ReassignmentMappings `json:"reassignment_mappings,omitempty"`
+	DestRoleSetKey       *string                     `json:"dest_role_set_key"`
```
kalafut: Should this have `omitempty`? The vast majority of optional params do have it, so if this is an exception, then a comment explaining why would be good.

NicolasLopes7: missed it, updating.

kalafut (approved):

PR #488, opened 2025-12-17
feat: add organization creation defaults

gabrielmeloc22:
Adds the organization creation defaults struct.


NicolasLopes7 (approved):
Can we add a test on the instance_settings client test just to make sure everything is serializing correctly? other than that LGTM

(also let's hold to merge until the full migration is complete)

PR #493, opened 2026-01-21
audit_logs change eventTime to int64

brunol95:
Updates eventTime to int64 according to changes in https://github.com/clerk/clerk_go/pull/16019

USER-4278

### DIFF
--- a/audit_logs.go
+++ b/audit_logs.go
@@ -1,12 +1,10 @@
 package clerk
 
-import "time"
-
 type AuditLog struct {
 	APIResource
 	Object          string         `json:"object"`
 	ID              string         `json:"id"`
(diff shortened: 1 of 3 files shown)


mauricioabreu (approved):
It's going to break users that expect `event_time` to be `time.Time` right? But since Golang is a compiled language, I think it does not do any harm

brunol95:
> It's going to break users that expect `event_time` to be `time.Time` right? But since Golang is a compiled language, I think it does not do any harm

@mauricioabreu this endpoint has no consumers yet as we are early in development as part of [proj-system-logs](https://clerkinc.slack.com/archives/C09LVRUBHR8)

mauricioabreu:
@brunol95 I was not aware of it. Good to know! Thank you

PR #499, opened 2026-01-29
feat: audit logs add actor and traceID query params

brunol95:
Part of USER-4450


BrandonRomano (approved):

audit_logs/client.go
```diff
@@ -36,6 +36,10 @@ type ListParams struct {
 	EndingBefore *string `json:"ending_before,omitempty"`
 	// Filter audit logs by subject (user ID or organization ID).
```
dmoerner: Just noting for future reference that right now our main subject is actually `sign_up_id`

dmoerner (approved):

audit_logs/client.go
BrandonRomano: Oh yeah; this subject can be anything!

PR #509, opened 2026-02-08
feat: add impersonation field to  audit logs response

brunol95:
Supports functionality added in https://github.com/clerk/clerk_go/pull/16537

### DIFF
--- a/audit_logs.go
+++ b/audit_logs.go
@@ -2,19 +2,20 @@ package clerk
 
 type AuditLog struct {
 	APIResource
-	Object       string          `json:"object"`
-	ID           string          `json:"id"`
-	EventTime    int64           `json:"event_time"`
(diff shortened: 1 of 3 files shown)


Jibaru (approved):
lgtm

dmoerner (approved):
I'm guessing that using a struct here instead of a `*string` is a forward-looking design since we might add more to the impersonator information? If so, looks good :)

brunol95:
> I'm guessing that using a struct here instead of a `*string` is a forward-looking design since we might add more to the impersonator information? If so, looks good :)

@dmoerner yeah that was my idea! When creating an actor token for impersonation the only required field is the user_id but we could definitely provide more data in the future. For example I could see a `reason` field where the impersonator has to give an explanation on why impersonation is required

PR #510, opened 2026-02-09
feat: add client_id and impersonator to audit logs query params

brunol95:
- [x] https://github.com/clerk/clerk_go/pull/16602
Part of USER-4550


Jibaru (approved):
lgtm. The only thing I'm not sure about is whether we should expose that information, but I think it was going to be improved later on.

brunol95:
> lgtm. The only thing I'm not sure about is whether we should expose that information, but I think it was going to be improved later on.

Yeah, the conclusion was that we are going to expose impersonator information but make it apparent in the UI that someone was impersonating!

BrandonRomano (approved):

PR #511, opened 2026-02-09
feat(oauth2idp): add `access_token_ttl` property to PATCH OAuthApplication

jfoshee:
So that DAPI can read/patch the OAuth Access Token lifetime.

The `access_token_ttl` property in the PATCH request will be optional and nullable:

- Omitting `access_token_ttl` means "do not modify"
- Specifying `access_token_ttl: null` means "reset to default" (i.e. set to null in the DB)

SDK consumers would set the parameters like this:

```
	params := &UpdateParams{
		Name: clerk.String("Test Application"),
		// Omit access token TTL, so unchanged
	}
```

```
	params := &UpdateParams{
		// Specify null to reset to default
		AccessTokenTTL: &clerk.OptionalNullableInt64{Value: nil},
	}
```

```
	updatedTTL := int64(3600)
	params := &UpdateParams{
		// Set to 1 hour
		AccessTokenTTL: &clerk.OptionalNullableInt64{Value: &updatedTTL},
	}
```

We have used this pattern in APIs before, but this seems to be the first time adding an optional in the Go SDK.

https://github.com/clerk/clerk_go/blob/b2fcf5076451c8a83ad7e4fbe3ebd23d2fcbb30b/pkg/json/int.go#L11-L15

Part of [USER-4657](https://linear.app/clerk/issue/USER-4657/add-bapidapi-support-to-update-oauth-access-token-ttl)

### DIFF
--- a/clerk.go
+++ b/clerk.go
@@ -521,3 +521,36 @@ func Bool(v bool) *bool {
 func JSONRawMessage(v json.RawMessage) *json.RawMessage {
 	return &v
 }
+
+// OptionalNullableInt64 represents an optional nullable int64 for PATCH semantics:
+// nil (omit) = do not change; value with Value=nil = reset to default (null); value with Value set = use that number.
+// Use as a pointer field with json:",omitempty" so the key is omitted when nil.
+type OptionalNullableInt64 struct {
+	Value *int64
+}
+
+// Interface implementations
+var _ json.Marshaler = (*OptionalNullableInt64)(nil)
+var _ json.Unmarshaler = (*OptionalNullableInt64)(nil)
+
+// MarshalJSON implements json.Marshaler. When Value is nil outputs null; otherwise outputs the number.
+func (o OptionalNullableInt64) MarshalJSON() ([]byte, error) {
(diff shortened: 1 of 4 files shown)


clerk.go
```diff
@@ -521,3 +521,36 @@ func Bool(v bool) *bool {
...
 }
+
+// OptionalNullableInt64 represents an optional nullable int64 for PATCH semantics:
+// nil (omit) = do not change; value with Value=nil = reset to default (null); value with Value set = use that number.
+// Use as a pointer field with json:",omitempty" so the key is omitted when nil.
+type OptionalNullableInt64 struct {
+	Value *int64
+}
+
+// Interface implementations
```
dmoerner, line 532: Good check

dmoerner (approved):
Switching between nil and null is a bit of a nightmare to reason about but this looks good to me!

jfoshee:
> Switching between nil and null is a bit of a nightmare to reason about but this looks good to me!

I can't argue with that. 😭 @dmoerner

PR #513, opened 2026-02-13
feat(organization-domain): Add `domains` query to list endpoint

LauraBeatris:


NicolasLopes7 (approved):
just the tests on the `organizationdomain/client_test.go` but otherwise lgtm

PR #516, opened 2026-02-13
fix(oauth2idp): round-trip serialization of optional int64 type

jfoshee:
so that it works as expected when DAPI relays the request to BAPI.

The code is slightly easier to understand, but for better or worse, it includes a code clone of an optional type.

See https://github.com/clerk/clerk_go/pull/16769 for the integration test

### DIFF
--- a/optional/int64.go
+++ b/optional/int64.go
@@ -1,0 +1,63 @@
+// Credits
+// https://www.calhoun.io/how-to-determine-if-a-json-key-has-been-set-to-null-or-not-provided/
+
+package optional
+
+import (
+	"encoding/json"
+	"reflect"
+)
+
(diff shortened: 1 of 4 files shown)


optional/int64.go
jfoshee, line 1: This is a direct code clone from the clerk_go repo. 
It doesn't seem desirable to _move_ optional types from clerk_go to the sdk repo because the optional types are not only used by DAPI. So I cloned it. :-/  It might be better to use a 3rd module for this in the long run. Or perhaps we won't need the ability to "reset to default" in this case anyway and the optional type could be removed.

dmoerner (approved):
Using a package here so it's `optional.Int64` instead of just `Int64` is perfect, nice improvement.

PR #525, opened 2026-03-04
feat(billing): add membership_limit_set_by to Organization response

mauricioabreu:


mzhong9723:
We decided to add this to BAPI?

paddycarver:
The DAPI endpoint calls BAPI to get the organization info; I suggested we add this to BAPI instead of mixing a database query with a BAPI query, for consistency and to avoid double-querying for the entitlement.

paddycarver (approved):

PR #527, opened 2026-03-05
feat(audit_logs): add offset param and total_count to list response

iagodahlem:
Add Offset field to ListParams to support offset-based pagination as an alternative to cursor-based pagination. When provided, the API returns results starting at the given offset position.

Add TotalCount field to AuditLogList response, populated when offset-based pagination is used, to allow clients to know the total number of matching records.

### DIFF
--- a/audit_logs.go
+++ b/audit_logs.go
@@ -20,8 +20,9 @@ type AuditLog struct {
 
 type AuditLogList struct {
 	APIResource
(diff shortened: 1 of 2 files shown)


BrandonRomano (changes requested):
Cannot support cursor based pagination on this endpoint.

PR #528, opened 2026-03-09
feat: support scim group role mapping

gabrielmeloc22:
Adds support for managing scim group role mapping


kalafut:
Needs `go generate .` for `api.go`?

scimgrouprolemapping/client.go
```diff
@@ -0,0 +1,114 @@
...
+
+// mappingList is a custom type to implement ResponseReader for array responses.
+type mappingList []*clerk.SCIMGroupRoleMapping
+
+func (*mappingList) Read(_ *clerk.APIResponse) {}
+
+// groupList is a custom type to implement ResponseReader for array responses.
+type groupList []*clerk.SCIMGroup
+
+func (*groupList) Read(_ *clerk.APIResponse) {}
```
NicolasLopes7: Goes slightly different than the pattern we usually write on this SDK but i'm not against it, just raising

NicolasLopes7 (approved):
lgtm

scimgrouprolemapping/client.go
gabrielmeloc22: i've tweaked it a bit, should be more cohesive with the codebase now

PR #536, opened 2026-04-02
audit_logs: update pagination cursor 

brunol95:
Matches changes made in https://github.com/clerk/clerk_go/pull/17925 to support new pagination style in audit logs. 

also fixes https://linear.app/clerk/issue/USER-4807/flaky-test-race-in-testpackagerevoke-testpackageverify-in-clerk-sdk-go

### DIFF
--- a/apikey/api_test.go
+++ b/apikey/api_test.go
@@ -13,6 +13,9 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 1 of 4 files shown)


jfoshee (approved):
I guess we don't have tparallel linter on this project... At any rate it would be good to have a comment somewhere, like at the top of the file, that briefly explains why these can't be parallel.

PR #537, opened 2026-04-09
feat: use canonical pagination on scim groups list

gabrielmeloc22:
Adds support for standard cursor pagination on SCIM groups list method


kalafut (approved):
LGTM but please at a PR description, even just a basic one.

PR #550, opened 2026-05-01
feat: Add instance communication settings

dmoerner:
Support GET and PATCH endpoints for the instance communication settings. Currently the backend only supports SMS blocklists.

### DIFF
--- a/instancesettings/client.go
+++ b/instancesettings/client.go
@@ -100,6 +100,39 @@ func (c *Client) UpdateRestrictions(ctx context.Context, params *UpdateRestricti
 	return instanceRestrictions, err
 }
 
+type UpdateCommunicationParams struct {
+	clerk.APIParams
+	// BlockedCountryCodes is the list of ISO-3166-1 alpha-2 country codes
+	// for which SMS messages will not be delivered. Pass an empty slice to
+	// clear the blocklist.
+	BlockedCountryCodes *[]string `json:"blocked_country_codes,omitempty"`
+}
+
+// GetCommunication retrieves the communication settings of the instance.
+func (c *Client) GetCommunication(ctx context.Context) (*clerk.InstanceCommunication, error) {
+	path, err := clerk.JoinPath(path, "/communication")
+	if err != nil {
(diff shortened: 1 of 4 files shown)


instancesettings/client.go
```diff
@@ -100,6 +100,39 @@ func (c *Client) UpdateRestrictions(ctx context.Context, params *UpdateRestricti
 	return instanceRestrictions, err
 }
 
+type UpdateCommunicationParams struct {
+	clerk.APIParams
+	// BlockedCountryCodes is the list of ISO-3166-1 alpha-2 country codes
+	// for which SMS messages will not be delivered. Pass an empty slice to
+	// clear the blocklist.
+	BlockedCountryCodes *[]string `json:"blocked_country_codes,omitempty"`
```
Grundlefleck, line 108: ❓ Are we deliberately less strict than on the backend?

There is a validate directive, (`validate:"omitempty,max=250,unique,dive,iso3166_1_alpha2`) on the server side. 

IMO it's better to be permissive on the client, since it's much easier to update and roll out a change to server-side validation than client side SDK validation. But I wanted to check that's how we roll with the go sdk.

instancesettings/client.go
```diff
@@ -100,6 +100,39 @@ func (c *Client) UpdateRestrictions(ctx context.Context, params *UpdateRestricti
...
+}
+
+// GetCommunication retrieves the communication settings of the instance.
```
Grundlefleck, line 111: 🙃 might just be me, but I find these comments to be a net-negative, taking up space to repeat the same information from the method signature.

But I guess that's just convention, regardless?

Grundlefleck (approved):
Couple of questions, but LGTM.

instancesettings/client.go
dmoerner: Yeah, as far as I know we leave this validation for the backend.

instancesettings/client.go
dmoerner: 🙃 In this case I agree but it is the convention.

PR #555, opened 2026-05-08
audit_logs: support OR-ing list filters via FilterMatch

BrandonRomano:
Adds a FilterMatch field on audit_logs.ListParams that controls how Subject, Type, Actor, TraceID, ClientID, and ImpersonatorUserID are combined when more than one is supplied:

  - clerk.AuditLogFilterMatchAll (default, also when nil): every supplied filter must match (AND). Preserves existing behavior.
  - clerk.AuditLogFilterMatchAny: an event matches if at least one of the supplied filters matches (OR).

EventTimeAfter, EventTimeBefore, and EndUserFacingOnly are always ANDed with the (possibly OR-ed) string filter group regardless of FilterMatch.

Merge after https://github.com/clerk/clerk_go/pull/18613


zourzouvillys (approved):

Jibaru (approved):

thiskevinwang (approved):
unblocking ✅

PR #557, opened 2026-05-15
feat(saml-connections): expose multi_valued on custom attributes

NicolasLopes7:
## Summary

Adds `MultiValued *bool` to `clerk.CustomAttribute` so SAML connection custom attributes can opt into multi-value mapping (multiple `<saml:AttributeValue>` children flowing into `public_metadata` as an array instead of being flattened to the first value).

Round-trips through both:
- `samlconnection` package — adds `CustomAttributes *[]clerk.CustomAttribute` to `CreateParams` (already on `UpdateParams` in v3).
- `enterpriseconnection` package — adds `CustomAttributes` to `CreateParamsSaml` for parity with `UpdateParamsSaml`.

## Why

Multi-value `<saml:Attribute>` is SAML-spec compliant and emitted by common IdPs (e.g. Okta groups, Azure AD roles). The current behavior flattens to the first value, which makes group/role mapping unreliable. Backend support is being added in parallel; this PR surfaces the field in the public SDK so consumers can read/write `multi_valued` per attribute.

## Changes

- `clerk.CustomAttribute`: new `MultiValued *bool` field (JSON `multi_valued,omitempty`)
- `samlconnection.CreateParams.CustomAttributes` (new)
- `enterpriseconnection.CreateParamsSaml.CustomAttributes` (new)
- `sdkVersion` -> `v3.1.0`
- CHANGELOG entry

## Backwards compatibility

- Pointer field, omitempty — existing callers unaffected.
- Field absent on response = nil = current behavior (single-value, first-only mapping).

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./...` — 398 tests, 45 packages, all pass
- [x] New tests: create + update round-trip with `multi_valued: true` and `multi_valued: false`, both `samlconnection` and `enterpriseconnection`

### DIFF
--- a/enterpriseconnection/client.go
+++ b/enterpriseconnection/client.go
@@ -42,6 +42,8 @@ type CreateParamsSaml struct {
 	IdpMetadata      *string                 `json:"idp_metadata,omitempty"`
 	AttributeMapping *AttributeMappingParams `json:"attribute_mapping,omitempty"`
 	ForceAuthn       *bool                   `json:"force_authn,omitempty"`
+	// CustomAttributes is an Experimental feature, not available for all customers.
+	CustomAttributes *[]clerk.CustomAttribute `json:"custom_attributes,omitempty"`
 }
 
 // CreateParamsOidc is the request body for creating an OIDC enterprise connection.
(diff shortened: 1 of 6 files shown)


jfoshee (approved):
It wasn't obvious to me what "multi-valued" meant, but I did confirm it is an accepted concept in SAML (as the PR says).

PR #558, opened 2026-05-15
remove metadata fields in user and organization Update method

brunol95:
This PR introduces a breaking change to the `user.Update` and `organization.Update` method removing metadata fields to align with the latest API contract.

- [x] update api version


BrandonRomano (approved):
Good call to make the change here 👍

PR #559, opened 2026-05-19
deprecate metadata fields in user and organization update method

brunol95:
This PR deprecates metadata fields on the user  and organization client `Update` method in favor of `ReplaceMetadata` or `UpdateMetadata`

- [x] update api version - 2026-05-12 is the official latest version

### DIFF
--- a/user/client.go
+++ b/user/client.go
@@ -84,29 +84,35 @@ func (c *Client) Get(ctx context.Context, id string) (*clerk.User, error) {
 
 type UpdateParams struct {
 	clerk.APIParams
-	FirstName                        *string          `json:"first_name,omitempty"`
-	LastName                         *string          `json:"last_name,omitempty"`
-	PrimaryEmailAddressID            *string          `json:"primary_email_address_id,omitempty"`
-	NotifyPrimaryEmailAddressChanged *bool            `json:"notify_primary_email_address_changed,omitempty"`
-	PrimaryPhoneNumberID             *string          `json:"primary_phone_number_id,omitempty"`
-	PrimaryWeb3WalletID              *string          `json:"primary_web3_wallet_id,omitempty"`
-	Username                         *string          `json:"username,omitempty"`
-	ProfileImageID                   *string          `json:"profile_image_id,omitempty"`
-	ProfileImage                     *string          `json:"profile_image,omitempty"`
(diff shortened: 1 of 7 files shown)


user/client.go
BrandonRomano, line 1: We have an upgrading file we will want to update here that we'll want to update https://github.com/clerk/clerk-sdk-go/blob/v3/UPGRADING.md with this PR - trying to keep all the breaking changes documented as we make them so we don't have to remember when we cut a release.

BrandonRomano (approved):
Preemptively approving - just be sure to add to the upgrading docs before merging!

user/client.go
brunol95: I updated the `UPGRADING.md` in the v3 branch [pull request](https://github.com/clerk/clerk-sdk-go/pull/558/changes#diff-f411a5009131db5a6c7242b49ca78488f2d274ef9e5e5f4de9262a42108a5386) as that includes the breaking change of removing the fields. For this PR, we are marking the fields as deprecated  but keeping the method backwards compatible. Let me know if this is the right process!

BrandonRomano: OH, this is v2, we *cannot* introduce a breaking change here

BrandonRomano: I see we're just tagging deprecated - that is fine!

BrandonRomano: Let me doubly check things here 👀

user/client.go
```diff
@@ -117,7 +123,41 @@ type UpdateParams struct {
 }
 
 // Update updates a user.
+//
+// If any of the deprecated metadata fields on UpdateParams is set, the
+// SDK splits the call: non-metadata fields are sent via PATCH /users/{id}
+// and the metadata fields are then sent via PUT /users/{id}/metadata (the dedicated metadata endpoint).
+// When only metadata fields are set, the PATCH call is skipped. The returned user
+// reflects the response of the final call made.
 func (c *Client) Update(ctx context.Context, id string, params *UpdateParams) (*clerk.User, error) {
```
BrandonRomano, line 132: There's slightly some risk that the user ends up half updated here (one request succeeds, one fails).  I think you're handling the errors correctly here though if that happens so the user knows there is an error so they retry.

BrandonRomano (approved):
Just re-approving for explicitness - this is g2g!

PR #565, opened 2026-06-04
fix(domain): include secondary flag in create params

Railly:
## Summary

Adds `is_secondary` to `domain.CreateParams` in the v3 SDK.

## Why

`clerk_go`'s Dashboard domain-create pass-through needs to preserve the secondary-domain flag when creating the first custom domain after a provider domain. BAPI already accepts the field; this SDK change keeps the request shape intact.

## Validation

- `go test ./domain -count=1`


Railly:
Closing this after Kevin's feedback. We do not want to expose `is_secondary` on the SDK/API surface for this flow. clerk_go#19322 has been reworked to pin the existing DAPI create + promote behavior without a new param.

PR #581, opened 2026-07-29
feat(user): add trusted device management

seanperez29:
Adds typed trusted-device responses and user client methods for listing active trusted devices and revoking an enrolled device. These methods support the clerk_go Dashboard API proxy while keeping Backend API as the canonical owner of trusted-device management.

Tests cover request paths, HTTP methods, list decoding, and revoked-device responses.

Clerk Go dependency: clerk/clerk_go#20204


user/client.go
```diff
@@ -571,6 +571,39 @@ func (c *Client) DeletePasskey(ctx context.Context, userID, identificationID str
...
+	}
+	req := clerk.NewAPIRequest(http.MethodGet, path)
+	resource := &trustedDeviceList{}
+	err = c.Backend.Call(ctx, req, resource)
+	return []*clerk.TrustedDevice(*resource), err
+}
+
+// Custom type needed in order to store the GET /v1/users/{userID}/trusted_devices
+// response array.
+type trustedDeviceList []*clerk.TrustedDevice
```
dmoerner: I think the proper way to do this is to make a list type that also embeds `APIResource`. See how we implement `UserList` in `user.go`.

dmoerner (approved):

PR #583, opened 2026-08-03
fix(fva): Fail closed on omitted fva

dmoerner:
Some Clerk JWTs don't include a factor verification age (`fva`). When we unmarshal these JWTs into our struct, they get the zero value of [0, 0]. In this case, the zero value in unsafe: It means both first and second factor were reverified within the last minute.

Add custom handling the unmarshaling code to explicitly unmarshal the `fva`. If the claim is not included, set it to the safe sentinel value of [-1, -1] instead of the unsafe zero value.

### DIFF
--- a/jwt.go
+++ b/jwt.go
@@ -74,9 +74,9 @@ func (s *SessionClaims) HasRole(role string) bool {
 func (s *SessionClaims) NeedsReverification(policy SessionReverificationPolicy) bool {
 	firstFactorAgeMinutes, secondFactorAgeMinutes := s.FactorVerificationAge[0], s.FactorVerificationAge[1]
 
-	firstFactorNeedsReverification := firstFactorAgeMinutes == -1 ||
+	firstFactorNeedsReverification := firstFactorAgeMinutes == factorVerificationAgeNotEnabled ||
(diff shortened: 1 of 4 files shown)


jwt.go
```diff
@@ -156,6 +156,49 @@ type Claims struct {
...
+	// the token or from Go. A slice can: nil means the claim wasn't there.
+	var raw struct {
+		FactorVerificationAge []int64 `json:"fva"`
+	}
+	if err := json.Unmarshal(data, &raw); err != nil {
+		return err
+	}
+
+	c.FactorVerificationAge = [2]int64{factorVerificationAgeNotEnabled, factorVerificationAgeNotEnabled}
+	if len(raw.FactorVerificationAge) == 2 &&
```
mauricioabreu, line 188: I think this is the first time I've seen business logic in this repo. Not saying it's wrong, just calling it out

mauricioabreu (approved):

jwt.go
dmoerner: It already leaked here, arguably `-1` as the first factor sentinel is not even part of our business logic, but it's already been encoded in the JS and Go SDKs, so it seems here to stay.

PR #588, opened 2026-08-11
fix(http): Scope JWK cache entries per Clerk instance

dominic-clerk:
The package-level JWK cache was keyed on the bare `kid`, so a key fetched for one Clerk instance was served to every other instance in the same process. Because a Clerk `kid` is the instance id, a token minted by instance B was a direct cache hit when presented to instance A, and `isValidIssuer` only checks for a Clerk-shaped issuer — so B's session claims were returned as an authenticated request in A's context.

Cache entries are now scoped by a hash of the Backend's API URL and secret key, exposed through the new `clerk.CacheScoper` interface that the default Backend implements. A cross-instance lookup misses and forces a fetch under the caller's own credentials. A custom Backend that does not implement `CacheScoper` cannot identify its instance, so its keys are not cached at all.

Fixes SDK-149


dmoerner (approved):
Looks correct to me.

PR #600, opened 2026-08-28
chore: rename audit_logs package to logs

brunol95:


Jibaru (commented):
I think this is a breaking change, right?

brunol95:
> I think this is a breaking change, right?

@Jibaru Yes this change is scoped only to `v3` which tolerates breaking changes

Jibaru (approved):

dmoerner (approved):

